PR #8, opened 2021-11-27
Enumerate adapters

gmodarelli:
Hello Michal. I've been playing with the repository, and I'm having a lot of fun. Thanks for sharing it with us all.

My system comes with both an AMD and an NVIDIA RTX card. At the moment, the graphics library automatically picks the first available adapter.
I've extended the code so that now the graphics library picks an adapter based on the following rules:
- must not be flagged as software adapter (DXGI_ADAPTER_FLAG_SOFTWARE)
- has the most amount of dedicated video memory
- can be used to create a D3D12Device

In the process of adding that code, I've fixed a minor typo and introduced more DGXI error codes and flags 6f97053.

I've also split this PR into two commits because although the code for picking a D3D12-compatible adapter works just fine 
0c29f83, the rules I've used are arbitrary, and I'm not sure about the zig code style I've used. I've only just started with zig :)

I also think that the `const S_FALSE` I've introduced in `graphics.zig` should be part of the `win32` library, just like `S_OK` is.

I'll be happy to apply any feedback you have on this :)

Thanks once more for open-sourcing this project!


michal-z:
Hi and thanks for the PR! I'm glad that you like the project a have some fun with it.

Generally PR looks fine, but for selecting the adapter I would like to use ```EnumAdapterByGpuPreference()``` with ```DXGI_GPU_PREFERENCE_HIGH_PERFORMANCE```. This method has been added in ```IDXGIFactory6``` interface so we would have to add it. Please let me know if you want to modify your code to use this method. For an example you can look here: https://github.com/microsoft/DirectML/blob/master/Samples/DirectMLSuperResolution/DeviceResources.cpp#L614

Thanks!

gmodarelli:
Hey Michal. That sounds like a good idea. I will try to add `IDXGIFactory6` and its functions tomorrow :)

Thanks for the feedback!

michal-z:
We can create ```IDXGIFactory6``` instead of ```IDXGIFactory1``` right away here: https://github.com/michal-z/zig-gamedev/blob/main/libs/common/graphics.zig#L109

```IDXGIFactory6``` is supported from Windows 10 Redstone 4 (1803) - we require at least Redstone 5 (1809) so it will alwyas work.

**Comment by gmodarelli** [comment]
Hello Michal, I've managed to implement the changes and add support for `IDXGIFactory6` and now we're using `EnumAdapterByGpuPerformance` to pick an adapter :)

Let me know if you have any further feedback.
Have a nice day!

File: libs/win32/dxgi.zig [hunk 1]
```diff
@@ -1179,6 +1185,57 @@ pub const IFactory5 = extern struct {
     }
 };
 
+pub const DXGI_GPU_PREFERENCE = UINT;
+pub const DXGI_GPU_PREFERENCE_UNSPECIFIED = 0;
+pub const DXGI_GPU_PREFERENCE_MINIMUM = 1;
+pub const DXGI_GPU_PREFERENCE_HIGH_PERFORMANCE = 2;
+
```
michal-z, line 1192: We should omit DXGI_ prefix here.

libs/win32/dxgi.zig
```diff
@@ -987,6 +987,12 @@ pub const IDevice = extern struct {
     }
 };
 
+pub const DXGI_ADAPTER_FLAGS = UINT;
+pub const DXGI_ADAPTER_FLAG_NONE = 0;
+pub const DXGI_ADAPTER_FLAG_REMOTE = 0x1;
+pub const DXGI_ADAPTER_FLAG_SOFTWARE = 0x2;
+pub const DXGI_ADAPTER_FLAG_FORCE_DWORD = 0x4;
+
```
michal-z, line 995: Again we don't need DXGI_ prefix.

libs/common/graphics.zig
```diff
@@ -150,14 +140,57 @@ pub const GraphicsContext = struct {
             }
         }
 
+        const suitable_adapter = blk: {
+            var adapter: ?*dxgi.IAdapter1 = null;
+
+            var adapter_index: u32 = 0;
+            var optional_adapter1: ?*dxgi.IAdapter1 = undefined;
```
michal-z, line 147: Let's init it to null innstead of undefined.

libs/common/graphics.zig
```diff
@@ -150,14 +140,57 @@ pub const GraphicsContext = struct {
...
         const device = blk: {
             var device: *d3d12.IDevice9 = undefined;
-            const hr = d3d12.D3D12CreateDevice(
-                null,
-                .FL_11_1,
-                &d3d12.IID_IDevice9,
-                @ptrCast(*?*c_void, &device),
-            );
+            const hr = innerblk: {
+                if (suitable_adapter) |adapter| {
```
michal-z, line 177: I think that we can simplify the code to this:
```
           const device = blk: {
                    var device: *d3d12.IDevice9 = undefined;
                    break :blk d3d12.D3D12CreateDevice(
                        if (suitable_adapter) |adapter| @ptrCast(*w.IUnknown, adapter) else null,
                        .FL_11_1,
                        &d3d12.IID_IDevice9,
                        @ptrCast(*?*c_void, &device),
                    );
            };
```

gmodarelli: Yes, that looks much better. Doing it now :)

PR #12, opened 2021-12-12
Makes PIX work with the latest zig version

gmodarelli:
Hello there, some small fixes to make samples compile and run when PIX is enabled.


michal-z:
Great, thanks!

PR #13, opened 2021-12-20
Versioned Root Signature

gmodarelli:
Hello Michal. This PR adds support for serializing versioned root signatures. It also adds all the missing structs, enums, flags and functions that I could find.

I've tested the new code by creating a sample root signature with a couple of 32-bit constants, a constant buffer and a descriptor table for bindless textures. Let me know if you want me to provide the sample code.

Have a nice day!


michal-z:
Very nice, thanks!

Let's make ```ROOT_SIGNATURE_VERSION``` an enum, we can just ignore ```ROOT_SIGNATURE_VERSION_1``` (we will force users to use ROOT_SIGNATURE_VERSION_1_x variants).

If you think that your code sample is complete we could include it to the package as an example.

gmodarelli:
The code sample is not ready since I've added the versioned root signature to my "main" project. I'll extract it into its own sample in the upcoming days and file a new PR if that's okay with you?

michal-z:
> The code sample is not ready since I've added the versioned root signature to my "main" project. I'll extract it into its own sample in the upcoming days and file a new PR if that's okay with you?

Sure, sounds good!

PR #15, opened 2021-12-30
Bindless textures sample

gmodarelli:
Hello there, I've finally understood how to work with bindless textures (or any shader visible descriptor for that matter) and I've put together a small sample to illustrate how it could work in the context of `zig-gamedev`.

I've had to make 2 main changes to `graphics.zig`, both non-breaking hopefully.

515e43f adds another `createGraphicsShaderPipeline*` variation that takes in a `Root Signature` as input.
36994cd adds a `PersistentDescriptorHeap` that can be reserved upon initialization and that I'm using in the provided sample for all bindless resources.

I think that the `PersistentDescriptorHeap` should be more integrated into `GraphicsContext` or maybe even collapsed into `DescriptorHeap`.

I've used Matt Pettineo's chapter and code from Ray Tracing Gems 2 as a guide for this implementation. The `DescriptorHeap` struct in particular (and associated operations) were really insightful.
https://github.com/TheRealMJP/DXRPathTracer/blob/master/SampleFramework12/v1.02/Graphics/GraphicsTypes.h#L38

I consider this PR an MVP or proof of concept. The API's I've added are not really ergonomic, "bindless mipmaps" are broken and the `PersistentDescriptorHeap` is not thread-safe. 

Before I go ahead, I'd like to have your thoughts and feedback since I feel like the changes I'm proposing are more "architectural" than a simple API usage sample :)

Have a great day!


michal-z:
Great, thanks! I will look at it later today or tomorrow morning.

michal-z:
Sorry for the delay. I got sick today - now I feel bit better but still not great.

I really like the direction but I would like to simplify few things.
Defining root signature in code is verbose this is why I always define this as string in the shader. What do you think about this approach?

I think we can quite easily add support for persistent descriptors directly to GraphicsContext. I have an early idea but I need to think it through a bit more.

Also, we now have some common mesh loading code in library.zig ```appendMeshPrimitive``` - can you please use this? You can see an example in ```mesh_shader_test.zig``` (just skip the meshoptimizer steps).

I really appreciate your work - it will make the framework better and more roboust!

gmodarelli:
Hey there, sorry to hear you're sick. I hope you get better soon! Thanks for making time for reviewing this nonetheless :)

> Defining root signature in code is verbose this is why I always define this as string in the shader. What do you think about this approach?

Yes, I've noticed how verbose it got while creating the sample. I don't have much experience with D3D12, so I trust your approach. I'll try to `#define` the root signature in the shader like in all the other samples. Should I remove the new function that takes in a root signature when creating a PSO?

> I think we can quite easily add support for persistent descriptors directly to GraphicsContext. I have an early idea but I need to think it through a bit more.

That's great! I had initially tried to change `GraphicsContext` myself to add support for it, but I was touching too many parts and I wasn't confident my changes weren't breaking things.

> Also, we now have some common mesh loading code in library.zig appendMeshPrimitive - can you please use this? You can see an example in mesh_shader_test.zig (just skip the meshoptimizer steps).

Absolutely, will change this right away!

> I really appreciate your work - it will make the framework better and more robust!

Thank you <3

michal-z:
Great, so much less code :) I'm looking into persistent descriptors right now.

gmodarelli:
Awesome! I am really looking forward to your solution!

michal-z:
I have submitted my code, usage is as follows:

```
// create resource
const cpu_handle = grfx.allocateCpuDescriptors(.CBV_SRV_UAV, 1);
// create shader resource vie using cpu_handle
const stable_index = grfx.allocatePersistentGpuDescriptors(1, cpu_handle);
```
```stable_index``` is an index into bound descriptor heap and can be directly passed to shader. Please let me know if it works for you.

**Comment by michal-z** [comment]
I think I found even better solution - please do not use current code just yet.

**Comment by michal-z** [comment]
You can now do:

```
const  descriptor = grfx.allocatePersistentGpuDescriptors(1);

pub const PersistentDescriptor = struct {
    cpu_handle: d3d12.CPU_DESCRIPTOR_HANDLE,
    gpu_handle: d3d12.GPU_DESCRIPTOR_HANDLE,
    index: u32,
};
```

**Comment by gmodarelli** [comment]
Really nice API! I will update the sample and add the missing shading parts as well (BRDF, LUT, etc...).

**Comment by gmodarelli** [comment]
Hello there, I've managed to replicate the PBR sample in my bindless sample.

Unfortunately, there are 2 things I'm struggling with that I can't get to work correctly:

1 - Mipmap generation doesn't seem to work when I create textures as bindless resources
2 - I cannot figure out how to generate Irradiance, Prefiltered Env and BRDF integration textures as bindless resources

In the case of mipmap generation, all mip levels, except level 0, are completely black.

**Comment by michal-z** [comment]
Nice! Will review the code tomorrow and try to help with those two issues. Thanks!

**Comment by michal-z** [comment]
All looks fine to me. I have only one suggestion. Instead of declaring an array of Textures2D we could use built in ```ResourceDescriptorHeap``` array (HLSL 6.6). This would allow us to remove this line:

```       
// Bind bindless texture descriptor table
grfx.cmdlist.SetGraphicsRootDescriptorTable(3, demo.bindless_descriptor_gpu_start);
```

And we would access textures as:
```
    Texture2D srv_ao_texture = ResourceDescriptorHeap[draw_const.ao_index];
```

What do you think?

gmodarelli:
Great, I will switch to `ResourceDescriptorHeap` :)

gmodarelli:
It worked like a charm, thanks for reviewing the PR :)

PR #18, opened 2022-02-05
Define d3d11 and dxgi return codes

hazeycode:
As defined at here [here](https://docs.microsoft.com/en-us/windows/win32/direct3d11/d3d11-graphics-reference-returnvalues) and [here](https://docs.microsoft.com/en-us/windows/win32/direct3ddxgi/dxgi-error) respectively.

I'm using these to turn HRESULTS into zig errors in a project that is only using the win32 lib from this repo. Also, it seems that the COM documentation defines a slightly different set of result codes than the above docs - hence I have left the return codes defined in misc.zig untouched.

On this topic, I'd like to move my error handling code into the win32 library. Would you be open to such changes? (it would be an alternative or potential eplacement to the error handling code in the common lib)

### DIFF
--- a/libs/win32/d3d11.zig
+++ b/libs/win32/d3d11.zig
@@ -3,6 +3,7 @@ const IUnknown = windows.IUnknown;
 const UINT = windows.UINT;
 const WINAPI = windows.WINAPI;
 const GUID = windows.GUID;
+const HRESULT = windows.HRESULT;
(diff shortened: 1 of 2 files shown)


michal-z:
Thanks for the PR!

Yes, I think that moving error handling to win32 library is a good idea.

PR #21, opened 2022-02-07
d3d11 bindings for vs and ps creation

hazeycode:
Adds the nessesary bindings to create pixel and vertex shaders on d3d11.

Also adds naughty bindings for d3dcompiler


michal-z:
Thanks!

Would be awesome to also have bindings for dxc (https://github.com/microsoft/DirectXShaderCompiler/blob/master/include/dxc/dxcapi.h). I plan to do it in the future (I will need it for hot shader recompilation) - but PR is very welcome :)

hazeycode:
Cool! Yeah dxc would better.
I will give you a heads up if I get to dxc bindings, so we don't duplicate effort there :)

PR #24, opened 2022-02-09
win32.d3d11 more bindings (2)

hazeycode:
adds d3d11 bindings relating to:
- binding shaders
- binding input layouts
- mapping/unmapping buffers
- setting render targets
- setting rasterizer state

### DIFF
--- a/libs/win32/d3d11.zig
+++ b/libs/win32/d3d11.zig
@@ -8,6 +8,10 @@ const HINSTANCE = windows.HINSTANCE;
 const SIZE_T = windows.SIZE_T;
 const LPCSTR = windows.LPCSTR;
 const FLOAT = windows.FLOAT;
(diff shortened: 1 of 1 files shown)


michal-z:
@hazeycode Just noticed that we could use `PRIMITIVE_TOPOLOGY` from d3dcommon.zig, right?

hazeycode:
@michal-z you're correct. I missed this remark in the d3d11 docs:

> The D3D11_PRIMITIVE_TOPOLOGY enumeration is type defined in the D3D11.h header file as a [D3D_PRIMITIVE_TOPOLOGY](https://msdn.microsoft.com/en-us/library/ff728726(v=vs.85)) enumeration, which is fully defined in the D3DCommon.h header file.

michal-z:
Yes, please and thanks!

PR #30, opened 2022-02-17
zmath: add Matrix functions

hazeycode:
- identity
- orthographicLh
- orthographicRh


michal-z:
Thanks!
One thing, orthographic proj actually allows `near < 0` you can define a cube that spans for example [-100, 100] range on all 3 axes. And I think that `near > far` also should be allowed to do inverse Z.
I will merge this and remove first assert `orthographicLh` and `orthographicRh`. Let me know if there is a problem with this.

PR #53, opened 2022-04-30
fix build errors for zig 0.10.0-dev.2023+032c722d2

garettbass:
I just installed zig on my M1 MacBook Pro, using homebrew, `brew install zig --HEAD`, and ended up with `zig version 0.10.0-dev.2023+032c722d2`, which reported a few minor compile errors when running `zig build` in the repository root.

- Apparently `std.ChildProcess.init()` no longer returns an error, so the `try` keyword is no longer relevant, but the error message was somewhat confusing to a newb like myself:
```
./build.zig:129:44: error: expected error union type, found 'std.child_process.ChildProcess'
    const child = try std.ChildProcess.init(&.{ "git", "submodule", "update", "--init", "--recursive" }, allocator);
                                           ^
```

- When the child process is declared `const child`, the next line, `child.cwd = thisDir();` becomes an error:
```
./build.zig:130:24: error: cannot assign to constant
    child.cwd = thisDir();
                       ^
```

- It seems `math.fabs()` has been replaced by the `@fabs()` builtin:
```
./libs/zmath/src/zmath.zig:3442:25: error: container 'std.math' has no member called 'fabs'
    var temp: f32 = math.fabs(angle);
                        ^
```

After fixing these, I was able to successfully build the root of the repo.  I'm not yet able to build `samples/triangle_wgpu` due to the following error:
```
./build.zig:6:17: error: import of file outside package path: '../../build.zig'
const Options = @import("../../build.zig").Options;
                ^
```

If I figure out how to fix that, I will create another PR.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -126,7 +126,7 @@ fn ensureSubmodules(allocator: std.mem.Allocator) !void {
     if (std.process.getEnvVarOwned(allocator, "NO_ENSURE_SUBMODULES")) |no_ensure_submodules| {
         if (std.mem.eql(u8, no_ensure_submodules, "true")) return;
     } else |_| {}
-    const child = try std.ChildProcess.init(&.{ "git", "submodule", "update", "--init", "--recursive" }, allocator);
+    var child = std.ChildProcess.init(&.{ "git", "submodule", "update", "--init", "--recursive" }, allocator);
     child.cwd = thisDir();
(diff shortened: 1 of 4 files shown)


garettbass:
I figured out how to successfully build and run `samples/triangle_wgpu`, and it works great on my M1 MacBook :)

michal-z:
Thanks for those fixes!

PR #54, opened 2022-05-03
Added fibers support to ztracy

briangold:
Hello Michal - this PR adds support for the Fiber annotations added to Tracy v0.8.0 and above. The actual runtime code is straightforward, but it requires an additional compilation flag (`-DTRACY_FIBERS`) is passed through, so I've added an options struct to the build step. 

According to the Tracy docs this has some small overhead, so I've left it off by default.

With this patch, an async event loop can now mark zones that span suspend points, and Tracy will keep the necessary context to show what the async task (aka fiber / coroutine / green-thread) was doing as though it was a single logical flow of control.

### DIFF
--- a/libs/zd3d12/README.md
+++ b/libs/zd3d12/README.md
@@ -52,7 +52,7 @@ pub fn build(b: *std.build.Builder) void {
     exe.addPackage(zd3d12.getPkg(b, options_pkg));
     exe.addPackage(zwin32.pkg);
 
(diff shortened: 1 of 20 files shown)


michal-z:
Thanks for the PR!
Will merge it tomorrow.

briangold:
I realized I hadn't changed any of the sample code link options, so the first commit would break things. I've pushed a second commit that addresses this, so please try it out and let me know if something looks wrong.

PR #56, opened 2022-05-12
zgpu: Fix warning for Git LFS

damienfir:
This fixes a slight instruction mistake when downloading the Git LFS files after the initial `git clone`.

(Nice work on the libraries by the way, it looks neat)


michal-z:
Thanks!

PR #57, opened 2022-05-15
zpool v0.1

garettbass:
This is my first attempt at a generic Pool, configurable for the number of bits per index and cycle, a specific resource type, and a set of columns of data that can be retrieved using a handle.  I wanted to try this approach because I suspect that the data access patterns for, a resource pointer and the rest of the resource meta-data may be different, such that there is greater information density when the two are stored in separate arrays.

Feedback is very welcome.


michal-z:
Amazing work! Many thanks! I would like to merge it right away.

Only two minor things, can you run `zig fmt` so that we have canonical form? I can do it if you want.
There is a dedicated allocator for tests: `std.testing.allocator` so let's use it.

Also, would be awesome to have a short readme file for `zpool` would you like to write it as next PR?

garettbass:
Thanks for the feedback.  I have made the suggested changes to the code and would be happy to make a separate PR to add a readme.

I am also interested in writing a future PR to replace zgpu's `ResourcePool` with zpool's `Pool`, because I'm curious to see what friction arises so I can make any necessary changes to streamline usage of `Pool`.

PR #61, opened 2022-05-23
WIP Attempt to use zpool in zgpu

garettbass:
Hello Michal.  I'm having trouble figuring out how to add a dependency from `zgpu` -> `zpool`, so that I can replace `zgpu`'s `ResourcePool` with `zpool.Pool`.  
This PR shows what I've tried so far.  I added a test to the end of zgpu.zig where I'm trying to import zpool, but when I run `zig test 
libs/zgpu/src/zgpu.zig`, I get the following error message:

```sh
$ zig test libs/zgpu/src/zgpu.zig 
./libs/zgpu/src/zgpu.zig:1721:19: error: unable to find 'zpool'
    const zpool = @import("zpool");
                  ^
```

Any advice you can offer would be greatly appreciated.  Thanks in advance!

### DIFF
--- a/libs/zgpu/build.zig
+++ b/libs/zgpu/build.zig
@@ -1,6 +1,7 @@
 const std = @import("std");
 const glfw = @import("../mach-glfw/build.zig");
 const gpu_dawn = @import("../mach-gpu-dawn/build.zig");
+const zpool = @import("../zpool/build.zig");
 
 pub const Options = struct {
     glfw_options: glfw.Options = .{},
@@ -17,6 +18,7 @@ fn buildLibrary(
     lib.setTarget(exe.target);
     glfw.link(exe.builder, lib, options.glfw_options);
     gpu_dawn.link(exe.builder, lib, options.gpu_dawn_options);
+    zpool.link(lib);
 
     // If you don't want to use 'dear imgui' library just comment out below lines and do not use `zgpu.gui`
     // namespace in your code.
@@ -33,6 +35,7 @@ fn buildLibrary(
     // If you don't want to use 'stb_image' library just comment out below line and do not use `zgpu.stbi`
     // namespace in your code.
     lib.addCSourceFile(thisDir() ++ "/libs/stb/stb_image.c", &.{"-std=c99"});
+    lib.addPackage(zpool.pkg);
 
     return lib;
 }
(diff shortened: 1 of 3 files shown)


libs/zpool/build.zig
```diff
@@ -1,10 +1,8 @@
 const std = @import("std");
 
-const main = .{ .zig = thisDir() ++ "/src/main.zig" };
```
garettbass, line 3: I thought perhaps things weren't working because I tried to be clever here, but changing this does not seem to have any effect.

libs/zgpu/build.zig
```diff
@@ -1,6 +1,7 @@
 const std = @import("std");
 const glfw = @import("../mach-glfw/build.zig");
 const gpu_dawn = @import("../mach-gpu-dawn/build.zig");
+const zpool = @import("../zpool/build.zig");
```
garettbass, line 4: Just trying to follow the pattern here and below, but I'm clearly missing something important.

libs/zgpu/src/zgpu.zig
```diff
@@ -1716,3 +1716,9 @@ pub const bglBuffer = gpu.BindGroupLayout.Entry.buffer;
 pub const bglTexture = gpu.BindGroupLayout.Entry.texture;
 pub const bglSampler = gpu.BindGroupLayout.Entry.sampler;
 pub const bglStorageTexture = gpu.BindGroupLayout.Entry.storageTexture;
+
+test "use zpool package from zgpu?" {
+    const zpool = @import("zpool");
```
garettbass, line 1721: Just added this test as a quick way to see whether the `zpool` package was accessible here.

```sh
$ zig test ./libs/zgpu/src/zgpu.zig 
./libs/zgpu/src/zgpu.zig:1721:19: error: unable to find 'zpool'
    const zpool = @import("zpool");
```

hazeycode: `zig test` doesn't invoke the calls to `addPackage`, you can use the `--pkg-begin` and `--pkg-end` options to specify packages (see `zig test --help`)

hazeycode: `zig test --pkg-begin zpool ./libs/zpool/src/zpool.zig --pkg-end ./libs/zgpu/src/zgpu.zig` should do it

hazeycode:
zig test doesn't invoke build.zig, you can use the --pkg-begin and --pkg-end options to specify packages (see zig test --help)

`zig test --pkg-begin zpool ./libs/zpool/src/zpool.zig --pkg-end ./libs/zgpu/src/zgpu.zig` should do it

michal-z:
Alternatively, you could add `zgpu` tests to our root `build.zig` and then run `zig build test` (this will run all tests).

`const zgpu_tests = @import("libs/zgpu/build.zig").buildTests(b, options.build_mode, options.target);`

garettbass:
Thank you both for the context on this issue.

PR #65, opened 2022-05-27
Change std.build.Pkg.path to std.build.Pkg.source

kamidev:
Fix breaking change in Zig master, see ziglang/zig#11557


michal-z:
Many thanks!

PR #67, opened 2022-05-29
Add zgpu dependency on zpool

garettbass:
- Set `-Dzgpu-dawn-from-source=true` on macOS
- Change obsolete `Pkg.path` to new `Pkg.source` in `zpool/build.zig`
- First naive attempt to add `zgpu` dependency on `zpool`.  This does not work, but I don't understand how it should work.  Please advise.

### DIFF
--- a/libs/zgpu/src/zgpu.zig
+++ b/libs/zgpu/src/zgpu.zig
@@ -1313,8 +1313,58 @@ fn ResourcePool(comptime Info: type, comptime Resource: type) type {
 
         fn addResource(self: *Self, gctx: GraphicsContext, info: Info) Handle {
             assert(info.gpuobj != null);
-            _ = gctx;
-            return self.pool.addAssumeNotFull(.{ .info = info });
+
+            if (self.pool.addIfNotFull(.{ .info = info })) |handle| {
+                return handle;
+            }
+
+            // If pool is free, attempt to remove a resource that is now invalid
+            // because of dependent resources which have become invalid.
+            // For example, texture view becomes invalid when parent texture
+            // is destroyed.
+            //
+            // TODO: We could instead store a linked list in Info to track
+            // dependencies.  The parent resource could "point" to the first
+            // dependent resource, and each dependent resource could "point" to
+            // the parent and the prev/next dependent resources of the same
+            // type (perhaps using handles instead of pointers).
+            // When a parent resource is destroyed, we could traverse that list
+            // to destroy dependent resources, and when a dependent resource
+            // is destroyed, we can remove it from the doubly-linked list.
+            //
+            // pub const TextureInfo = struct {
+            //     ...
+            //     // note generic name:
+            //     first_dependent_handle: TextureViewHandle = .{}
+            // };
+            //
+            // pub const TextureViewInfo = struct {
+            //     ...
+            //     // note generic names:
+            //     parent_handle: TextureHandle = .{},
+            //     prev_dependent_handle: TextureViewHandle,
+            //     next_dependent_handle: TextureViewHandle,
+            // };
+            if (self.removeResourceIfInvalid(gctx)) {
+                if (self.pool.addIfNotFull(.{ .info = info })) |handle| {
+                    return handle;
+                }
+            }
+
+            // TODO: For now we just assert if pool is full - make it more roboust.
+            assert(false);
+            return Handle.nil;
+        }
+
+        fn removeResourceIfInvalid(self: *Self, gctx: GraphicsContext) bool {
+            var live_handles = self.pool.liveHandles();
+            while (live_handles.next()) |live_handle| {
+                if (!gctx.isResourceValid(live_handle)) {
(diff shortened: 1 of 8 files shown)


michal-z:
Thanks!

`zgpu/build.zig` doesn't need to explicitly link with `zpool`, instead the user of the `zgpu` package needs to pass the dependencies.

The user of the `zgpu` package will need to do this:

```zig
const zgpu_options = zgpu.BuildOptionsStep.init(b, .{ .dawn = .{ .from_source = false } });
const zgpu_pkg = zgpu.getPkg(&.{ zgpu_options.getPkg(), glfw.pkg, zpool.pkg }); // deps of the zgpu package

exe.addPackage(zgpu_pkg);
exe.addPackage(glfw.pkg);
exe.addPackage(zpool.pkg); // only needed if an application uses zpool directly
```

michal-z:
I've submitted two first changes from this PR.

garettbass:
Can you rely on these import paths if the caller is providing these packages externally?  If they are required to place their versions of GLFW and Dawn in exactly these paths, don’t we already have enough information to call addPackage on their behalf?

```zig
const glfw = @import("../mach-glfw/build.zig");
const gpu_dawn = @import("../mach-gpu-dawn/build.zig");

pub const BuildOptions = struct {
    use_imgui: bool = true,
    use_stb_image: bool = true,

    glfw: glfw.Options = .{},
    dawn: gpu_dawn.Options = .{},
};
```

michal-z:
Notice that for `zpool` (and other our libs) you wouldn't add an import path to `zgpu/build.zig`. `glfw` and `dawn` are exceptions for now because those are externals and we don't have full control in this case (but I may actually remove those imports..).

The idea is: I don't want to have any imports in a `build.zig` for a given package. The user should provide the dependencies because the user knows where those packages are.

In other words, `zgpu` should be coupled with `zpool` only by `const zpool = @import("zpool");` placed in the source code. `zgpu/build.zig` should not hardcode `zpool` package path.

This is what I would like to achieve. Let me know what do you think.

garettbass:
I agree with the goal, just trying to figure out how to make it explicit at build time so the user doesn’t have to compile, then see that “zpool” package is not found to discover the missing dependency.

michal-z:
I think that it is enough to just document dependencies.

`zgpu` package depends on its options (`BuildOptionsStep.getPkg()`) and on `zpool` package. All those packages must be passed to `zgpu.getPkg()`.

michal-z:
I have moved `mach-glfw` and `mach-gpu-dawn` into `zgpu`. So that `zgpu` is a single package now which automatically links with `glfw`. Linking with `zgpu` looks:

```zig
const zgpu = @import("libs/zgpu/build.zig");

pub fn build(b: *std.build.Builder) void {
    ...
    const zgpu_options = zgpu.BuildOptionsStep.init(b, .{ .dawn = .{ .from_source = false } });
    const zgpu_pkg = zgpu.getPkg(&.{zgpu_options.getPkg()});

    exe.addPackage(zgpu_pkg);

    zgpu.link(exe, zgpu_options);
}
```

garettbass:
I have reimplemented `zgpu.ResourcePool` using `zpool.Pool`, and found that all of the `*_wgpu` demos appear to behave the same as before.

I did discover that the `bullet_physics_test_wgpu` demo shows only a black screen and an IMGUI.  I'm not sure what it should do, but it behaves the same for me on `main`.

michal-z:
Nice!

Black screen in `bullet_physics_test_wgpu` is normal, it is wip and nothing renders yet :)

michal-z:
Overall it looks very good thanks for your work!

One difference I see is that on `main` branch `addResource()` uses our custom `isResourceValid()` to check if we can re-use slots.
For example, `TextureViewHandle` stores a handle to its parent texture. So, if a parent texture becomes invalid (deleted) texture view handle also becomes invalid (it points to non-existent texture) and we can re-use its slot in the pool.

I think that this logic is missing in this PR. What do you think?

garettbass:
Thanks, I will have to take a closer look to understand that scenario and figure out how to reproduce it.

garettbass:
Interesting.  I should be able to do something similar by iterating over `pool.liveHandles()` in search of any that are now invalid due to dependencies.

Thanks for the context, my eyes glazed over when I looked at the implementation and I mistakenly assumed I knew what it was doing 😅

Thinking about when this should be done.  You currently scan for invalid resources anytime you need to add a new resource, and bail out as soon as you find an unused slot.  Since zpool uses a free queue to immediately locate an unused slot, I’m tempted to employ this “search and destroy” approach only when the pool is full.  That would mean that invalid resources may hang around in the pool longer, but I think that may be ok since, in general, resource owners should be releasing them as they go.

michal-z:
I think it should be perfectly fine to free them only when the pool is full.

garettbass:
I implemented the algorithm as we discussed and added some commentary about a possible alternative approach for future consideration.  I also rebased onto latest `main` to fix merge conflicts.

PR #68, opened 2022-05-31
Simplify handle.zig

garettbass:


libs/zpool/src/handle.zig
```diff
@@ -57,78 +57,71 @@ pub fn Handle(
...
+            var u = CompactUnion{ .id = self.id };
             return .{
-                .index = self.index(),
-                .cycle = self.cycle(),
+                .index = u.bits.index,
+                .cycle = u.bits.cycle,
             };
         }
 
-        pub fn index(self: Self) AddressableIndex {
```
garettbass, line 104: Favor operating on `AddressableHandle` in all cases where `index` and `cycle` are needed.  The goal is an API that discourages shifting and masking the `index` and `cycle` bits for every operation.

libs/zpool/src/handle.zig
```diff
@@ -57,78 +57,71 @@ pub fn Handle(
...
-            pub fn compact(handle: AddressableHandle) Compact {
-                const _index = @intCast(Index, handle.index);
-                const _cycle = @intCast(Cycle, handle.cycle);
-                return Compact.init(_index, _cycle);
-            }
-
-            pub fn eql(a: AddressableHandle, b: AddressableHandle) bool {
-                return a.index == b.index and a.cycle == b.cycle;
+            /// Returns the corresponding `Handle`
+            pub fn compact(handle: AddressableHandle) CompactHandle {
```
garettbass, line 121: Perhaps I should rename this method simply `handle()`

michal-z: Yes, I think we should go with handle().

michal-z:
Looks good! Can you please rebase to latest?

garettbass:
I have rebased and changed `AddressableHandle.compact()` to `AddressableHandle.handle()` as discussed.

PR #70, opened 2022-06-12
Add orthographicOffCenter matrix creations functions, 

mlarouche:
Those matrices are used frequently in 2D games to have traditional screen coordinates with the origin being at top-level, X and Y coordinates being always positive.


michal-z:
Thanks!

PR #71, opened 2022-06-19
zmath: Add benchmark results from Ryzen 7 5800X

hazeycode:
Thought it'd be interesting to compare. Let me know if you'd like me to recollect the results for this CPU at any point in the future.


michal-z:
Thanks!

PR #91, opened 2022-08-29
Add basic ImPlot functions

ckrowland:
For ImPlot::PlotLine and ImPlot::PlotBar I force a numeric type for the values, but the ImPlot function is generic


michal-z:
Thanks for the PR!

It will require some work before I can merge it. Do you want to clean it up a bit? I can do it if you want.

ckrowland:
Clean it up? Like the formatting or the function calls? I'm not sure what you have in mind.

michal-z:
There are few things:
- Hardcoded constants in: `zguiSetupLegend`, `zguiSetupLegend`, `zguiSetupAxes`
- Generally `zgui` passes args in structs which allows us to provide default values - we should follow this convention
- I think we need to to prefix ImPlot functions with `zguip` (or something) to avoid name clashes (for example,  there is ImGui::PushStyleColor() and ImPlot::PushStyleColor())

ckrowland:
Ok, yes thanks for the clarification. I can do that sometime this week.

michal-z:
Sure, no rush.

michal-z:
Looks good, thanks! Will merge it today.

PR #103, opened 2022-09-10
zglfw: Added fns to set input callbacks.

foxnne:
Hi! I just added the set callback functions that I was using prior to the recent change to glfw. I basically just added them as they were previously, but tried to follow what you were doing with the rest of it. Let me know if you'd rather avoid the callbacks, but from what I could tell, its the only way to get some input (for instance, mouse scrolling).


michal-z:
Hi and thanks for the PR!

PR #104, opened 2022-09-11
zwin32 stage3 compat

### LINKED ISSUE #102: zwin32 fails to build with latest Zig
zwin32 fails to build using the latest Zig compiler (0.10.0-dev.3900+ab4b26d8a)

fn ptr syntax has recently been changed.

hazeycode:
closes #102


michal-z:
Thanks! Looks good. I think we can remove support for stage1 (I'm already doing this in `zaudio`).

hazeycode:
Requires upstream std lib changes: https://github.com/ziglang/zig/pull/12813

hazeycode:
@michal-z I'm not sure about this change: https://github.com/michal-z/zig-gamedev/pull/104/commits/2523b456fdc658f24883b8f1366abd3580477396
What do you think?

michal-z:
Does below work?

```zig
pub const RAYTRACING_INSTANCE_DESC = extern struct {
    Transform: [3][4]FLOAT,
    InstanceIDAndMask: u32,
    bits: packed struct(u32) {
        InstanceContributionToHitGroupIndex: u24,
        Flags: u8,
    },
    AccelerationStructure: GPU_VIRTUAL_ADDRESS,
};
```

If not, let's go with your solution

hazeycode:
Yeah that works :+1: 
https://github.com/michal-z/zig-gamedev/commit/285a26dfbf0a9c57edcd5a1e6dde33d34260c4ff

PR #106, opened 2022-09-13
zgui: DrawList api refinement

Jack-Ji:


michal-z:
Hi, many thanks for your work!

I have incorporated changes from this PR in this commit: https://github.com/michal-z/zig-gamedev/commit/2bf04e5504592ec8540fdccd0cb88b03fc8848f1

I have slightly changed the API:
- All args are passed via struct (I think this makes reading the UI code easier)
- Arrays are passed to C code via pointers (this decreases number of args that we need to pass)

Also, I have added some example code to `gui_test_wgpu` sample app.

Let me know what you think.

Jack-Ji:
Hi, I agree with your choice, consistency is important in lib design. Although there's a minor bug need fixing, just created another pull request #107 .

PR #108, opened 2022-09-14
add Viewport support

meshula:
Hi, I found myself needing the `Viewport->WorkPos` and `->WorkSize` values, so I prototyped what the API might look like. This PR also has an unrelated demo (wrinkles) in it, sorry about the clutter.

I was wondering if I could request viewport functions as demonstrated here in zgui.cpp + zgui.zig?

I've marked the PR draft because I don't expect it to be merged, but feel free to cherry-pick if you like.

Thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 !*.hlsl
 !*.hlsli
 !*.m
+!*.wgsl
 !LICENSE*
 
 !*.ini
--- a/build.zig
+++ b/build.zig
@@ -28,6 +28,7 @@ pub fn build(b: *std.build.Builder) void {
         );
         // TODO: Re-enable zenet when stage3 compiler is more stable
         //installDemo(b, network_test.build(b, options), "network_test");
+        installDemo(b, wrinkles.build(b, options), "wrinkles");
     }
 
     //
@@ -187,6 +188,7 @@ const physically_based_rendering_wgpu = @import("samples/physically_based_render
 const bullet_physics_test_wgpu = @import("samples/bullet_physics_test_wgpu/build.zig");
 const audio_experiments_wgpu = @import("samples/audio_experiments_wgpu/build.zig");
 const gui_test_wgpu = @import("samples/gui_test_wgpu/build.zig");
+const wrinkles = @import("samples/wrinkles/build.zig");
 
 pub const Options = struct {
     build_mode: std.builtin.Mode,
--- a/libs/zgui/src/zgui.cpp
+++ b/libs/zgui/src/zgui.cpp
@@ -1641,3 +1641,19 @@ ZGUI_API void zguiPlot_PlotLineValues(const char* label_id, const int* values, i
 ZGUI_API void zguiPlot_EndPlot() {
     ImPlot::EndPlot();
 }
+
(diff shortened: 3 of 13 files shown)


michal-z:
Hi, and thanks for the patch!
It is now committed in: https://github.com/michal-z/zig-gamedev/commit/132d6763ff6ba4951ba9a171b373b1b3752b9172

michal-z:
@meshula Btw, can I ask what are you working on?

meshula:
I'm working with a partner on a new mathematical representation of time, to replace the core calculations in OpenTimelineIO. We've discovered a way to eliminate the need to round off or snap time to framerates, and to match simulations of varying framerates, and to get the same results for time calculations across applications. The demo is meant to visualize how it works for when we roll it out, because in a way it's "quaternions for time", and I'd like to have people think it's intuitive rather than a mystery bag of math. 

We're going to break opentime (https://github.com/AcademySoftwareFoundation/OpenTimelineIO/tree/main/src/opentime) out into its own library, written in C, so you can follow there over the next few months if you're interested, or in the "wrinkles (in time)" demo. I absolutely expect it to be super useful for games, and we're doing it in C so we can use it from all the languages we use :)

PS, I've built the demo inside of a fork of zig-gamedev because I'm struggling to build zig-gamedev when it is a git  submodule and my focus is getting my demo working, not solving the submodule problem ;). My working theory is that the comptime thisDir() function doesn't return the current build.zig's directory, but the root build.zig's directory. That's why did a spin of exploring the simulations repo to work out how it's being accomplished there.

meshula:
Fantastic work on zgui by the way. I have any number of my own partial wrappings for Dear Imgui and microui, depending on what I needed on any one project. zgui by far has the nicest wrapping pattern I've tried.

michal-z:
@meshula This sounds very interesting! I will be following your work for sure!

Personally I don't like submodules but below project got it working with zig-gamedev:
https://github.com/ckrowland/simulations

For now, the recommended way of using zig-gamedev is to just copy needed libraries or use the whole package as you did.

meshula:
I haven't tried copying before. I assume that would mean copying the z* libs in zig-out/libs to somewhere in my project, and then use LibExeObjStep.addPackage and addPackagePath to point to each of the libraries, which I could then @import in the normal way.

michal-z:
Yes, here is an instruction for `zgpu`:


Copy zgpu, zpool and zglfw folders to a libs subdirectory of the root of your project.

Then in your build.zig add:

```zig
const zgpu = @import("libs/zgpu/build.zig");
const zpool = @import("libs/zpool/build.zig");
const zglfw = @import("libs/zglfw/build.zig");

pub fn build(b: *std.build.Builder) void {
    ...
    const zgpu_pkg = zgpu.getPkg(&.{ zpool.pkg, zglfw.pkg });

    exe.addPackage(zgpu_pkg);
    exe.addPackage(zglfw.pkg);

    zgpu.link(exe);
    zglfw.link(exe);
}
```

meshula:
Perfect, thank you very much!!

meshula:
I see you have similar instructions in the individual lib readme files. I put together a new repo following that pattern, and it worked like a charm. https://github.com/meshula/wrinkles

PR #111, opened 2022-09-15
JoltC: Add more JPH::Body bindings

hazeycode:
Adds `GetInverseInteria` through `GetInverseCenterOfMassTransform` in order of API reference


michal-z:
This great, we can really improve the state of SW when working like this!

PR #113, opened 2022-09-17
JoltC: bindings wip

hazeycode:
Declares:
- JPH_AABox
- JPH_Plane
- JPH_Sphere
- JPH_MotionProperties

Defines bindings for
- JPH::MotionProperties

### DIFF
--- a/libs/zjolt/libs/JoltC/JoltC.h
+++ b/libs/zjolt/libs/JoltC/JoltC.h
@@ -114,7 +114,6 @@ typedef uint32_t JPH_SubShapeID;
 typedef uint32_t JPH_CollisionGroupID;
 typedef uint32_t JPH_CollisionSubGroupID;
 
-typedef struct JPH_PhysicsSystem    JPH_PhysicsSystem;
 typedef struct JPH_Shape            JPH_Shape;
 typedef struct JPH_PhysicsMaterial  JPH_PhysicsMaterial;
 typedef struct JPH_TempAllocator    JPH_TempAllocator;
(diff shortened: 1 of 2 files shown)


libs/zjolt/libs/JoltC/JoltC.h
```diff
@@ -151,13 +167,76 @@ typedef struct JPH_BroadPhaseLayerInterfaceVTable JPH_BroadPhaseLayerInterfaceVT
...
+
+    // 3rd cache line (least freqently used)
+    // 4 byte aligned
+    JPH_Sphere          sleep_test_spheres[3];
+    float               sleep_test_timer;
+
+#ifdef JPH_ENABLE_ASSERTS
+    JPH_MotionType      cached_motion_type;
+#endif
+};
```
hazeycode, line 238: I'm having some trouble to get the size of the struct to be correct. Any ideas @michal-z?

hazeycode: > I'm having some trouble to get the size of the struct to be correct. Any ideas @michal-z?

I think this is the last thing and then this PR is ready

hazeycode:
https://github.com/michal-z/zig-gamedev/pull/113#discussion_r973513816
i think this is the last thing before this PR is ready

michal-z:
Thanks, I will look at the `JPH_MotionProperties` size later today.

michal-z:
@hazeycode You missed one field :)
```
float                   mMaxAngularVelocity;
```

michal-z:
Actually, fields that are `Float3` can't be 4 element arrays, sorry. So, `force` and `torque` must have 3 elements.

PR #115, opened 2022-09-19
JoltC: fix struct member alignments

hazeycode:
I think we need to be more explicit with alignment of some members because of how we are defining vectors?


michal-z:
I have fixed few issues. Sizes and offsets are now correct.

We need to follow below rules:

`Float3 vector;` maps `float vector[3];`
`Vec3 vector;` maps `alignas(16) float vector[4];`

hazeycode:
Gotcha!

PR #117, opened 2022-09-21
zglfw/zgui: Added ability to set standard cursors and keep zgui from overriding.

foxnne:
I attempted to add the `setCursor` method which needed the `Cursor` struct to be created, as well as the `Shape` enum. I followed what was done on the previous wrapper, but maybe you don't want to go that route?

I believe I found an issue while trying this however: I believe this function doesn't work in conjunction with `zgui`, as when setting the cursor to a standard cursor, I get no errors, and see the cursor momentarily when the window loads, but after things are running I lose the ability to set the cursor. I know Dear Imgui has cursor functionality as well and I believe this may be overriding the `glfw` cursor?

So this may be a good way to test or a placeholder PR to serve as a reminder. I believe [this issue](https://github.com/ocornut/imgui/issues/1918) shows we may need to set some flags with zgui to get that functionality back?

EDIT: Found the issue to indeed be what was listed in the above issue, I added a way to set the config flags in `io` and now setting the config flags to `no_mouse_cursor_change` allows glfw to control the mouse cursor visuals normally. This may be something you want to set by default?

### DIFF
--- a/libs/zglfw/src/zglfw.zig
+++ b/libs/zglfw/src/zglfw.zig
@@ -53,6 +53,28 @@ pub const MouseButton = enum(i32) {
     eight,
 };
 
+pub const Cursor = extern struct {
+    ptr: *glfwCursor,
+    pub const glfwCursor = opaque {};
+    pub const Shape = enum(c_int) {
+        arrow = 0x00036001,
+        ibeam = 0x00036002,
+        crosshair = 0x00036003,
+        pointing_hand = 0x00036004,
+        resize_ew = 0x00036005,
+        resize_ns = 0x00036006,
+        resize_nwse = 0x00036007,
+        resize_nesw = 0x00036008,
+        resize_all = 0x00036009,
+    };
+
+    pub fn createStandard(shape: Shape) Cursor {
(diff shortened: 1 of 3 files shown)


michal-z:
Thanks for the PR!

Mouse cursor shape and visibility is automatically handled by our imgui GLFW backend. If you want to disable this default behavior you need to set `no_mouse_cursor_change` flag, this is expected. I think this is fine because most users that want to bother with setting custom cursors. What do you think?

I will tweak `Cursor` definition a bit (we want to use `*opaque` instead of `extern struct`.

michal-z:
I have added this functionality in below commits:

https://github.com/michal-z/zig-gamedev/commit/74381e0dc4b1114725221e9f6c3b6bdd703d7089
https://github.com/michal-z/zig-gamedev/commit/f553ec442ce854cb57896d7fbdda79647f6308ae

Please let me know if this is all you need, thanks!

foxnne:
Works for me! Thank you! I'll likely want to be able to create custom cursors at some point but for now this works, appreciate it!

PR #123, opened 2022-10-03
zjobs v0.1

garettbass:
This is a candidate for a new zig-gamedev library, `zjobs`.  It provides a generic `JobQueue` implementation that can schedule work to be completed on background threads.  Dependencies between jobs are supported, so you can schedule a job that will not begin until some other job or jobs have completed.  A simple example is provided as a test and in the README, and documentation for each of the public functions in `JobQueue` provides more detailed usage information.


michal-z:
Thanks! Great job :)

PR #124, opened 2022-10-03
changed job entry point from 'main' to 'exec'

garettbass:


garettbass:
rebased and pushed again to clean up history so there is only a single non-merge commit in the PR

Issue #41, opened 2022-03-28
Instanced line rendering demo - nice and fast lines in 3D


michal-z:
https://wwwtyro.net/2019/11/18/instanced-lines.html

PR #133, opened 2022-10-10
Instanced pills sample

Pyrolistical:
Implemented the instanced pill in 2d as a sample from #41 

This depends on #131 and #132. Will rebase once those are merged.


michal-z:
Very nice work! Many thanks for the contribution.

Few minor issue:

It is better to declare buffer handles as:
```zig
vertex_buffer: zgpu.BufferHandle = .{},
```
This declares an invalid handle that does not point to any resource yet. All functions that receive invalid handles are just skipped. So, to re-create a buffer you can write (you don't need to check if buffer is valid):
```zig
gctx.destroyResource(demo.vertex_buffer); // will be skipped internally if buffer is invalid
demovertex_buffer = // create new buffer
```

You don't need to destroy resources in `deinit()` - graphics context will do it for you.

Please stick to naming convention `funcName()` (camel for function names).

Pyrolistical:
Thank you for the feedback.

I've fixed the issues you have laid out.

Please give any other feedback you have. I want to learn how to best structure apps in zig/webgpu/zig-gamedev.

michal-z:
`instance_pills_wgpu.zig` file should be named `InstancePillsWgpu.zig` this is a zig convention for files that includes a "direct content of a struct" all other files should be named lower-case as you currently have.

michal-z:
@Pyrolistical I did some minor changes:

* Code formatting changes to break long lines
* It is better to use `zm.f32x4()` to init `zmath` vectors
* I realized that it is cleaner to explicitly declare `DemoState` struct

Hope you are fine with those changes.

Tomorrow I will announce new sample app on Twitter stating that you are the author of course.

Once again, thanks for the contribution and waiting for more ;)

Pyrolistical:
Of course. I just want that is best for clarity to the reader.

I'll stick with explicit struct in the future.

Why is it better to use `zm.f32x4()` instead of `.{ ... }` literal?

michal-z:
Using `zm.f32x4()` forces user to pass 4 args. When using `.{}` it is possible to pass fewer (by mistake), for example:
```zig
const v = zm.F32x4{ 1.0, 2.0 };
```
Above code compiles but I'm not sure what value will be assigned to `z` and `w` components.

PR #136, opened 2022-10-12
Avokadoen's combine test case with some suggested fixes by Garett

garettbass:
Great work from @Avokadoen on the test case for `JobQueue.combine()`.  I chose a different resolution for the compile error they encountered, and restored a loop in `combine()` that allows it to accept more `prereqs` than will fit in a single `CombinePrereqsJob`.  All tests passing on `zig 0.10.0-dev.4280+c3d67c5c4`


Avokadoen:
I wonder if you could maybe remove the jobs = undefined? Just do no default initialization maybe?

garettbass:
@Avokadoen I made the change you suggested, thanks!

PR #138, opened 2022-10-13
WIP Layers sample

Pyrolistical:
Sample that shows how to organize passes into layers for 2D graphics. 

The key idea is to simplify definition of instances but without adding additional abstractions. For example the [definition of pill is _mostly_ in a single file](https://github.com/Pyrolistical/zig-gamedev/blob/layers/samples/layers_wgpu/src/pill.zig).

Has some idea in common with https://github.com/regl-project/regl and https://github.com/pex-gl/pex-context.

Some issues (not necessarily blockers):

1. Should [`graphics.zig`](https://github.com/Pyrolistical/zig-gamedev/blob/layers/samples/layers_wgpu/src/graphics.zig) exist?
2. How manual should the API be? Right now consumer needs to know to recreate vertex/index/instance buffer. Can be this be automatic?


michal-z:
Thanks for the PR! I will review it in coming days.

PR #139, opened 2022-10-15
zgui: Add 'addFontFromFileWithConfig'

### LINKED ISSUE #137: zgui: Add icon font support for text icons.
Small issue, but I was hoping to use font-awesome icons in zgui, [similar to how Prime31 did with zig-upaya](https://github.com/prime31/zig-upaya/blob/master/src/deps/imgui/font_awesome.zig). For any interface it would be nice to have access to icons. 

It looks like [here](https://github.com/prime31/zig-upaya/blob/b040acab13c7af00c3ce0eade03e1f3b0b1d5b02/src/upaya.zig#L222) is a good start.

foxnne:
Closes #137.

### DIFF
--- a/libs/zgui/src/gui.zig
+++ b/libs/zgui/src/gui.zig
@@ -87,11 +87,38 @@ pub const ConfigFlags = enum(u32) {
     is_touch_screen = 1 << 21,
 };
 
+pub const FontConfig = extern struct {
+    font_data: *u0 = undefined,
+    font_data_size: i32 = 0,
+    font_data_owned_by_atlas: bool = true,
+    font_no: i32 = 0,
+    size_pixels: f32 = 0.0,
+    oversample_h: i32 = 3,
+    oversample_v: i32 = 1,
+    pixel_snap_h: bool = false,
+    glyph_extra_spacing: [2]f32 = .{ 0.0, 0.0 },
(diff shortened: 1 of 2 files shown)


michal-z:
Thanks!

It would be better to initialize `ImFontConfig` using the constructor on the C++ side.

```zig
pub const ImFontConfig = struct {
    //...
    pub fn init() ImFontConfig {
        return zguiImFontConfig_Init();
    }
};
```
zgui.cpp:
```cpp
ImFontConfig zguiImFontConfig_Init(void) {
    return ImFontConfig();
}
```

This guarantees that the default values will be always the same (we won't have to track changes in newer versions of imgui).

Please use `Wchar` type for a `range` argument, it is defined as:
```zig
pub const Wchar = u16;
```

michal-z:
Thanks for the changes.

One more thing, we should allow `config` and `ranges` to be `null` because user may want to pass only one of them.

```zig
 addFontFromFileWithConfig("myfont.ttf", 16.0, null, ranges);
 addFontFromFileWithConfig("myfont.ttf", 16.0, config, null);
```

foxnne:
I apologize for being slow! The kids had a birthday party to go to and I was trying to get it done before but didn't make it. Thanks!

PR #142, opened 2022-10-19
Added msaa to layers sample

Pyrolistical:
I found it non-obvious on how to add msaa when there are multiple color attachments involved, so I added it to the layers sample.


Pyrolistical:
Hmm, `zgui` is hardcoded not multi-sample: https://github.com/michal-z/zig-gamedev/blob/main/libs/zgui/libs/imgui/imgui_impl_wgpu.cpp#L575

This means this samples cannot use zgui as is. How should we solve this? Should we render zgui separately?

michal-z:
Thanks for the PR!

I see two solutions:
* We could render zgui to the back buffer after the MSAA resolve, or
* We could pass the number of samples to `zgui.backend.init()`

Both should work.
Yes, I think we should pass the number of samples to `zgui.backend.init()` as you did above.

Pyrolistical:
oh ya derp. I'll just render directly to the backbuffer. I added a commit to show that.

I think adding samples to init would be an annoying breaking change.

PR #145, opened 2022-10-26
zstbi: Add `stbi_load_from_memory` and `Image.initFromData`

foxnne:
Allows initializing an image from a buffer and length in memory. The use case for me is reading the contents of an image within a zipped file using a zip library and then creating an image from it to write to a texture and draw with zgui. The only thing I wasn't sure about was HDR images. Should I allow the user to pass that in directly and handle it like the other init function?

### DIFF
--- a/libs/zstbi/src/zstbi.zig
+++ b/libs/zstbi/src/zstbi.zig
@@ -119,14 +119,14 @@ pub const Image = struct {
         };
     }
 
-    pub fn initFromData(buffer: []const u8, forced_num_channels: u32) !Image {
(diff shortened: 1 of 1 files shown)


foxnne:
[Here](https://github.com/foxnne/pixi/blob/0ead2ef4b6b843cbfb27c37ec28360d60cae38a0/src/editor/editor.zig#L96) is the use case if you have a better idea.

michal-z:
Thanks!

Looks fine, I will add HDR support later.

Let's make a slice from `buffer: [*]u8, len: usize` -> `data: []const u8`.

PR #148, opened 2022-11-01
zgflw: Joystick and Gamepad support

Pyrolistical:
Please be critical on how I exposed the glfw API.

This PR looks big but isn't. 1,797 lines are due to the included gamecontrollerdb.txt


michal-z:
Thanks for the PR!

I will review it tomorrow morning.

michal-z:
I would like to expose `Joystick` as an opaque pointer. `jid` can be just a `u4` we don't need an enum.

```zig
pub const Joystick = *opaque {
    pub fn getGuid(joystick: Joystick) [:0]const u8 {
        return std.mem.span(glfwGetJoystickGUID(@intCast(i32, @ptrToInt(joystick))).?);
    }
    extern fn glfwGetJoystickGUID(jid: i32) ?[*:0]const u8;

    pub fn getName(joystick: Joystick) [:0]const u8 {
        return std.mem.span(glfwGetJoystickName(@intCast(i32, @ptrToInt(joystick))).?);
    }
    extern fn glfwGetJoystickName(jid: i32) ?[*:0]const u8;
};

pub fn getJoystick(jid: u4) ?Joystick {
    if (isJoystickPresent(jid)) {
        return @intToPtr(Joystick, @intCast(usize, jid));
    }
    return null;
}

pub fn isJoystickPresent(jid: u4) bool {
    return glfwJoystickPresent(@intCast(i32, jid)) != 0;
}
extern fn glfwJoystickPresent(jid: i32) i32;
```

michal-z:
Or perhaps, it would be better to define it like:

```zig
pub const Joystick = struct {
    id: u4,

    pub fn getGuid(joystick: Joystick) [:0]const u8 {
        return std.mem.span(glfwGetJoystickGUID(@intCast(i32, @ptrToInt(joystick))).?);
    }
    extern fn glfwGetJoystickGUID(jid: i32) ?[*:0]const u8;

    pub fn getName(joystick: Joystick) [:0]const u8 {
        return std.mem.span(glfwGetJoystickName(@intCast(i32, @ptrToInt(joystick))).?);
    }
    extern fn glfwGetJoystickName(jid: i32) ?[*:0]const u8;
};
```

What do you think?

Pyrolistical:
I like the struct version. I've made the API easier to use by having init return an optional joystick and giving the ability to "upgrade" a joystick to a gamepad.

I've made those changes, take another look.

libs/zglfw/src/zglfw.zig
```diff
@@ -233,6 +244,147 @@ pub const CursorMode = enum(i32) {
...
+
+    pub fn init(jid: u4) ?Joystick {
+        const self: Joystick = .{ .jid = jid };
+        if (!self.present()) {
+            return null;
+        }
+        return self;
+    }
+
+    pub fn present(self: Joystick) bool {
```
michal-z, line 269: Let's call it `isPresent()`.

libs/zglfw/src/zglfw.zig
```diff
@@ -233,6 +244,147 @@ pub const CursorMode = enum(i32) {
...
+        return glfwJoystickPresent(@intCast(i32, self.jid)) == @boolToInt(true);
+    }
+    extern fn glfwJoystickPresent(jid: i32) i32;
+
+    pub fn getGUID(self: Joystick) [*:0]const u8 {
```
michal-z, line 274: Let's call it `getGuid()` (Zig convention). Return type should be `[:0]` (requires `std.mem.span()` in impl).

libs/zglfw/src/zglfw.zig
michal-z: Actually this should be a global function called `isJoystickPresent(jid: u4) bool`. It should not be possible to create an instance of `Joystick` if it is not present on the system.

libs/zglfw/src/zglfw.zig
```diff
@@ -233,6 +244,147 @@ pub const CursorMode = enum(i32) {
...
+    pub const ButtonAction = enum(u8) {
+        release = 0,
+        press = 1,
+    };
+
+    jid: u4,
+
+    pub const maximum_supported = 16;
+
+    pub fn init(jid: u4) ?Joystick {
```
michal-z, line 261: I would prefer a global function `getJoystick(jid: u4) ?Joystick`. `init()` function suggests that `deinit()` should be also present.

libs/zglfw/src/zglfw.zig
```diff
@@ -233,6 +244,147 @@ pub const CursorMode = enum(i32) {
...
+        return state[0..@intCast(usize, count)];
+    }
+    extern fn glfwGetJoystickAxes(jid: i32, count: *i32) [*]const f32;
+
+    const empty_buttons: [0]ButtonAction = undefined;
+    pub fn getButtons(self: Joystick) []const ButtonAction {
+        var count: i32 = undefined;
+        const state = glfwGetJoystickButtons(@intCast(i32, self.jid), &count);
+        if (count == 0) {
+            return &empty_buttons;
```
michal-z, line 295:
```suggestion
            @as([*]const ButtonAction, undefined)[0..0]
```

libs/zglfw/src/zglfw.zig
```diff
@@ -233,6 +244,147 @@ pub const CursorMode = enum(i32) {
...
+        return glfwGetJoystickGUID(@intCast(i32, self.jid));
+    }
+    extern fn glfwGetJoystickGUID(jid: i32) [*:0]const u8;
+
+    const empty_axes: [0]f32 = undefined;
+    pub fn getAxes(self: Joystick) []const f32 {
+        var count: i32 = undefined;
+        const state = glfwGetJoystickAxes(@intCast(i32, self.jid), &count);
+        if (count == 0) {
+            return &empty_axes;
```
michal-z, line 284:
```suggestion
            return @as([*]const f32, undefined)[0..0];
```

libs/zglfw/src/zglfw.zig
```diff
@@ -233,6 +244,147 @@ pub const CursorMode = enum(i32) {
...
+        return @ptrCast([]const ButtonAction, state[0..@intCast(usize, count)]);
+    }
+    extern fn glfwGetJoystickButtons(jid: i32, count: *i32) [*]const u8;
+
+    fn isGamepad(self: Joystick) bool {
+        return glfwJoystickIsGamepad(@intCast(i32, self.jid)) == @boolToInt(true);
+    }
+    extern fn glfwJoystickIsGamepad(jid: i32) i32;
+
+    pub fn gamepad(self: Joystick) ?Gamepad {
```
michal-z, line 306: Let's call it `asGamepad()`.

libs/zglfw/src/zglfw.zig
```diff
@@ -233,6 +244,147 @@ pub const CursorMode = enum(i32) {
...
+
+    pub fn init(jid: u4) ?Gamepad {
+        if (Joystick.init(jid)) |joystick| {
+            return joystick.gamepad();
+        } else {
+            return null;
+        }
+    }
+
+    pub fn getName(self: Gamepad) []u8 {
```
michal-z, line 367: Return type should be `[:0]`.

libs/zglfw/src/zglfw.zig
```diff
@@ -233,6 +244,147 @@ pub const CursorMode = enum(i32) {
...
+        var state: State = undefined;
+        _ = glfwGetGamepadState(@intCast(i32, self.jid), &state);
+        // return value of glfwGetGamepadState is ignored as
+        // it is expected this is guarded by glfwJoystickIsGamepad
+        return state;
+    }
+    extern fn glfwGetGamepadState(jid: i32, state: *Gamepad.State) i32;
+};
+
+pub fn updateGamepadMappings(mappings: [*:0]const u8) bool {
```
michal-z, line 382:
```suggestion
pub fn updateGamepadMappings(mappings: [:0]const u8) bool {
```

libs/zglfw/src/zglfw.zig
```diff
@@ -233,6 +244,147 @@ pub const CursorMode = enum(i32) {
...
+    };
+
+    pub const State = extern struct {
+        buttons: [15]Joystick.ButtonAction,
+        axes: [6]f32,
+    };
+
+    jid: u4,
+
+    pub fn init(jid: u4) ?Gamepad {
```
michal-z, line 359: `init()` suggests that the object is constructed and that `deinit()` should be called to destruct it. Let's remove this function.

```zig
if (zglfw.getJoystick(1)) |joystick| {
    if (joystick.asGamepad()) |gamepad| {
    }
}
```

michal-z (changes requested):
Thanks. I've added few comments.

Pyrolistical:
Thanks for the feedback! I've incorporated it all.

michal-z:
Looks good, thanks!

Pyrolistical:
Thanks, I would just like to reflect on the steps we took here and I think it may serve as a good template for future contributions.

1. Create strawman api/implementation and sample using it
2. Debate strawman to decide on better api/implementation design
3. Execute on better api/implementation design
4. Last review to nitpick final details such as code conventions and better types

michal-z:
Sounds good, should we publish those steps somewhere? WHat do you think?

Pyrolistical:
Let's do this again first

PR #153, opened 2022-11-06
Exposes mouse and keyboard input in zgui.io to allows a zig imgui backend to work

DeanoC:
Exposes imgui events for mouse, keyboard and focused events
Adds the keyboard enums and modifiers.

### DIFF
--- a/libs/zgui/src/gui.zig
+++ b/libs/zgui/src/gui.zig
@@ -190,6 +190,27 @@ pub const io = struct {
     /// `pub fnsetDeltaTime(delta_time: f32) void`
     pub const setDeltaTime = zguiIoSetDeltaTime;
     extern fn zguiIoSetDeltaTime(delta_time: f32) void;
+
+    pub const addFocusEvent = zguiIoAddFocusEvent;
+    extern fn zguiIoAddFocusEvent(focused: bool) void;
+
+    pub const addMousePositionEvent = zguiIoAddMousePositionEvent;
+    extern fn zguiIoAddMousePositionEvent(x: f32, y: f32) void;
+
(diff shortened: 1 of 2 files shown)


libs/zgui/src/gui.zig
```diff
@@ -207,7 +228,185 @@ pub const Font = *opaque {};
...
+    left_ctrl,
+    left_shift,
+    left_alt,
+    left_super,
+    right_ctrl,
+    right_shift,
+    right_alt,
+    right_super,
+    menu,
+    _0,
```
michal-z, line 258: Let's call those those fields `one`, `two`, etc. This will be consistent with what we did in `zglfw`.

libs/zgui/src/gui.zig
```diff
@@ -207,7 +228,185 @@ pub const Font = *opaque {};
...
+    gamepad_rstickdown,
+    gamepad_rstickleft,
+    gamepad_rstickright,
+
+    mod_ctrl,
+    mod_shift,
+    mod_alt,
+    mod_super,
+
+    _,
```
michal-z, line 370: This shouldn't be needed. Is there any specific reason to add it?

michal-z (changes requested):
Thanks!

Looks good, just two minor issues.

libs/zgui/src/gui.zig
DeanoC: In theory they key enum isn't complete, there can be values upto 512 that don't have a specific enum value. I don't know if thats needed anywhere but I add the _ in case its useful somewhere?

DeanoC: I'll remove it for now, it can always be added if we find a need

PR #158, opened 2022-11-13
Offscreen zgui text rendering and transform

Pyrolistical:
Discussed in https://github.com/michal-z/zig-gamedev/issues/95, here is the implementation of offscreen zgui text to texture and transformation.

I ran into quite a few issues and looking for a better design.

Issues:

1. Initially I interleaved `zgui.backend.newFrame` for the main UI and the offscreen, but this breaks imgui as it keeps track of data between frames in order for the text input and mouse drag focus to work. The solution to this is to create a second imgui context and switch between them.
2. Creating a second context for imgui was easy enough but `zgui.backend.init` assumes the backend is glfw, which the second context is not. The solution was to duplicate `backend_glfw_wgpu.zig` to `offscreen_wgpu.zig`
3. The next problem was both were relying on the same `imgui_impl_wgpu.cpp` which held static data. The solution to that was to duplicate `imgui_impl_wgpu.cpp` to `imgui_impl_offscreen_wgpu.cpp`

I added the ability to switch between the glfw and offscreen context with `useGlfw` and `useOffscreen`. I maintained backwards compatibility as by default `useGlfw` is implied. But once the offscreen context is enabled, things like` zglfw.pollEvents` need to be prefixed with `useGlfw`

As per trying out this process from last time:
- [x] Create strawman api/implementation and sample using it
- [ ] Debate strawman to decide on better api/implementation design
- [ ] Execute on better api/implementation design
- [ ] Last review to nitpick final details such as code conventions and better types

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -41,6 +41,7 @@ pub fn build(b: *std.build.Builder) void {
     installDemo(b, instanced_pills_wgpu.build(b, options), "instanced_pills_wgpu");
     installDemo(b, layers_wgpu.build(b, options), "layers_wgpu");
     installDemo(b, gamepad_wgpu.build(b, options), "gamepad_wgpu");
+    installDemo(b, text_transform_wgpu.build(b, options), "text_transform_wgpu");
 
     //
     // Tests
@@ -109,6 +110,7 @@ const gui_test_wgpu = @import("samples/gui_test_wgpu/build.zig");
 const instanced_pills_wgpu = @import("samples/instanced_pills_wgpu/build.zig");
 const layers_wgpu = @import("samples/layers_wgpu/build.zig");
 const gamepad_wgpu = @import("samples/gamepad_wgpu/build.zig");
+const text_transform_wgpu = @import("samples/text_transform_wgpu/build.zig");
 
 pub const Options = struct {
     build_mode: std.builtin.Mode,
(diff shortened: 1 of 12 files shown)


michal-z:
Many thanks for your work. I think this is a very valuable experiment. I think this approach is to complicated and cumbersome.

Looks like there is a simpler way:
https://github.com/ocornut/imgui/blob/master/imgui.h#L2581
https://github.com/michal-z/zig-gamedev/blob/main/libs/zgui/libs/imgui/imgui_impl_wgpu.cpp#L442

Drawing command can use callbacks for custom drawing code. We could use this callback to bind a custom render pipeline do the transformations. I will be investigating this approach some more.

Pyrolistical:
I don't understand how the callback applies to this situation. I see that is useful if one wanted to render a texture, but that doesn't solve the text rendering problem.

How are we going to render the text to a texture?

Instead of having to implement rendering offscreen with zgui for text rendering, we could use https://github.com/harfbuzz/harfbuzz

michal-z:
Using the callback we can draw `AddText()` command (from DrawList API) directly to a back buffer (no need for off-screen texture). In our custom drawing code we can use our own shader and transform text however we want.

Pyrolistical:
Ah, I see. That makes sense.

michal-z:
@Pyrolistical I hope this won't discourage you from making more PRs. You're doing great work!

Pyrolistical:
Oh don't worry about that. False starts are a good thing.

PR #160, opened 2022-11-19
zgui exposed more getters

Pyrolistical:
exposed

- ImGui::GetIO().DisplaySize
- ImGui::GetContentRegionMax
- ImGui::GetWindowContentRegionMin
- ImGui::GetWindowContentRegionMax


michal-z:
Thanks!

PR #170, opened 2022-12-13
zmath: Added util.zig.

Srekel:


libs/zmath/src/util.zig
```diff
@@ -0,0 +1,169 @@
...
+const zm = @import("zmath.zig");
+const std = @import("std");
+const math = std.math;
+const expect = std.testing.expect;
+
+pub fn matTranslation(m: zm.Mat) zm.Vec {
+    return m[3];
+}
+
+pub fn matForward(m: zm.Mat) zm.Vec {
```
michal-z, line 35: I think the better name would be just `getForwardVec()`. The `mat*()` function names are reserved for constructor-like functions that return a matrix (there are some exceptions though, like `identity()`, `rotationX()`).

libs/zmath/src/util.zig
```diff
@@ -0,0 +1,169 @@
...
+// angleMod(angle: f32) f32
+//
+// ==============================================================================
+
+const zm = @import("zmath.zig");
+const std = @import("std");
+const math = std.math;
+const expect = std.testing.expect;
+
+pub fn matTranslation(m: zm.Mat) zm.Vec {
```
michal-z, line 31: I think the better name would be just `getTranslationVec()`. The mat*() function names are reserved for constructor-like functions that return a matrix (there are some exceptions though, like identity(), rotationX()).

libs/zmath/src/util.zig
```diff
@@ -0,0 +1,169 @@
...
+    return m[3];
+}
+
+pub fn matForward(m: zm.Mat) zm.Vec {
+    return zm.normalize3(zm.mul(zm.f32x4(0.0, 0.0, 1.0, 0.0), m));
+}
+
+pub fn matUp(m: zm.Mat) zm.Vec {
```
michal-z, line 39: I think the better name would be just `getUpVec()`. The mat*() function names are reserved for constructor-like functions that return a matrix (there are some exceptions though, like identity(), rotationX()).

libs/zmath/src/util.zig
```diff
@@ -0,0 +1,169 @@
...
+    return zm.normalize3(zm.mul(zm.f32x4(0.0, 1.0, 0.0, 0.0), m));
+}
+
+pub fn matRight(m: zm.Mat) zm.Vec {
```
michal-z, line 43: I think the better name would be just `getRightVec()`. The mat*() function names are reserved for constructor-like functions that return a matrix (there are some exceptions though, like identity(), rotationX()).

libs/zmath/src/util.zig
```diff
@@ -0,0 +1,169 @@
...
+//
+// ==============================================================================
+
+const zm = @import("zmath.zig");
+const std = @import("std");
+const math = std.math;
+const expect = std.testing.expect;
+
+pub fn matTranslation(m: zm.Mat) zm.Vec {
+    return m[3];
```
michal-z, line 32: I think that we should set `w` component to 0 because user can take this vector and add to something and he may get unexpected result, for example:
```zig
const p0 = f32x4(1.0, 2.0, 3.0, 1.0); // point in space with w=1.0
const p1 = p0 + getTranslationVec(m); // now p1 has w==2.0 which is not expected
```

libs/zmath/src/util.zig
```diff
@@ -0,0 +1,169 @@
...
+}
+
+pub fn matForward(m: zm.Mat) zm.Vec {
+    return zm.normalize3(zm.mul(zm.f32x4(0.0, 0.0, 1.0, 0.0), m));
```
michal-z, line 36: I think it can be optimized to:
`zm.normalize3(zm.f32x4(m[0][2], m[1][2], m[2][2], 0.0));`

libs/zmath/src/util.zig
```diff
@@ -0,0 +1,169 @@
...
+}
+
+pub fn matUp(m: zm.Mat) zm.Vec {
+    return zm.normalize3(zm.mul(zm.f32x4(0.0, 1.0, 0.0, 0.0), m));
```
michal-z, line 40: I think it can be optimized to:
`zm.normalize3(zm.f32x4(m[0][1], m[1][1], m[2][1], 0.0));`

libs/zmath/src/util.zig
```diff
@@ -0,0 +1,169 @@
...
+}
+
+pub fn matRight(m: zm.Mat) zm.Vec {
+    return zm.normalize3(zm.mul(zm.f32x4(1.0, 0.0, 0.0, 0.0), m));
```
michal-z, line 44: I think it can be optimized to:
`zm.normalize3(zm.f32x4(m[0][0], m[1][0], m[2][0], 0.0));`

michal-z (changes requested):

michal-z:
Many thanks for PR! Few, things, mostly naming convention.

michal-z:
Awesome!

Just one note, we already have `zm.modAngle()` function, it works with scalars and vectors.

PR #173, opened 2022-12-17
zmesh: remove assert in remapVertexBuffer

Avokadoen:
_remapVertexBuffer_ incorrectly asserted that destination slice was bigger than source. It will pack vertices in a smaller destination
slice.


michal-z:
Thanks! I think we should remove this assert altogether.

Avokadoen:
Fixed

PR #176, opened 2022-12-27
zgui: expose api for managing custom DrawList

Jack-Ji:


michal-z:
Thanks! Looks good.

One question, it is not obvious why we need another render pass for new draw list in the demo, could you please explain?

Jack-Ji:
Hi michal, happy holiday!

About additional render pass, I'm not so sure actually, as I haven't  dipped my toes into zgpu yet. If you know better, please correct me.

Jack-Ji:
After some digging, a separate render pass is indeed uncalled for. However, the sample need to create separate pipline/vertex buffer/index buffer/bind group, is that correct? Do you think it's a bit too much for this sample?

PR #177, opened 2022-12-31
fix(zgui): add column arg to BeginTable and fix outer_size paramater type on zguiBeginTable

yusdacra:
I hit some compile errors trying to use `zgui.beginTable` and this change fixes it.


michal-z:
Thanks!

michal-z:
@yusdacra I did a follow-up commit which removes the default value for the `column`.

PR #178, opened 2022-12-31
fix(zgui): use correct name in beginTable, columns -> column

yusdacra:
Seems like the change that changed the name to `column` didn't change `args.columns` usage in `beingTable` itself


michal-z:
Yes you are right, thanks!

PR #179, opened 2023-01-01
fix(zmath): fix usages of thisDir for relative @src.file

Luexa:
At least on NixOS x86_64, `thisDir()` will return relative paths when invoked from ZLS due to `@src().file` no longer consistently being an absolute path (likely due to Zig stage1 relying on WASI). This actually leads to ZLS *crashing* entirely, so `zig-gamedev` needs to be updated to resolve project-relative paths with `cwd()`.

Although this solution seems a bit fragile and will be annoying to implement across the whole of `zig-gamedev`, it seems to work consistently and successfully allows ZLS to read build configuration properly again.

Note that for general usage of `zig build`, paths will be generated at comptime as before since `@src.file` appears to be absolute in those situations.

Also note that currently I have only fixed the issue for `zmath`, which is why I have created this PR as a draft.

### DIFF
--- a/libs/zmath/build.zig
+++ b/libs/zmath/build.zig
@@ -1,9 +1,17 @@
 const std = @import("std");
 
-pub const pkg = std.build.Pkg{
-    .name = "zmath",
-    .source = .{ .path = thisDir() ++ "/src/main.zig" },
-};
+pub fn pkg(b: *std.build.Builder) std.build.Pkg {
+    const cached_result = struct {
+        var pkg: ?std.build.Pkg = null;
+    };
+    if (cached_result.pkg == null) {
+        cached_result.pkg = .{
+            .name = "zmath",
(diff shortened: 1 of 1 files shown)


michal-z:
Thanks for the PR!

This complexity is a bit unfortunate, I really liked the simplicity of `thisDir()`. Would it be possible to refactor the code into a single function?

Also, I can't repro the issue on Ubuntu 22.04, do you think it is NixOS specific?

Luexa:
It's likely possible to refactor it into a single function, I may have
gotten a little carried away.

I doubt it's specific to NixOS, but it's probably difficult to consistently
reproduce. What I know for sure is that the Zig compiler can't guarantee
absolute paths considering it has to work from WASI.

Luexa:
Closing due to this behavior being worked around by ZLS (https://github.com/zigtools/zls/pull/898)

PR #182, opened 2023-01-03
zstbtt: stb_truetype bindings

hazeycode:
Work-in-progress stb_truetype bindings. Not ready for merge, opened for early feedback.


michal-z:
Thanks! Looks good.

One thing I would like to improve is removing the usage of `[*c]` - we should be using `[*]` or `[*:0]` (for C strings).

marler8997:
Just letting you know I recently converted a small TTF library "libschrift" to Zig in case you're interested in using it.  https://github.com/marler8997/libschrift

hazeycode:
> Just letting you know I recently converted a small TTF library "libschrift" to Zig in case you're interested in using it. https://github.com/marler8997/libschrift

Looks good. Thanks for the heads up!
Reasons for me choosing stb_truetype are that it's widely used / battle tested and that it's libc dependencies can be overidden (portable). Do you know what libschrift's story is in this regard?

marler8997:
Oh wow it's libc dependencies can be overridden? That's great. I don't know much about libschrift other than it was small and worked out of the box when I tested it and ported it. I might need need to look into stb myself

hazeycode:
> Oh wow it's libc dependencies can be overridden? That's great. I don't know much about libschrift other than it was small and worked out of the box when I tested it and ported it. I might need need to look into stb myself

It's true - the stb libs do a great job with this sort of thing.

Here we override malloc, free so that a zig allocator can be provided (as has been done with zstbi also). I have also overridden assert for both stb_truetype and stb_rect_pack (optional dep) which I needed to do to build for WASM.

image

hazeycode:
I'm not sure if I'll actually try to support wasm, I was just interested to see. But I think I only actually need this for an offline asset pipeline. Will likely leave removing libc dep as a todo.

marler8997:
Supporting wasm was actually the reason I decided to port libschrift to zig. I also ran into that asser_fail linker issue when it was in C. That and just wanting to understand how it all worked so it was worth it in the end.

hazeycode:
I've been trying out @marler8997 's schrift port and I really like it. I have decided to use it instead of stb_truetype.

So I'm probably not going to continue with these bindings. I will leave it in [this branch](https://github.com/hazeycode/zig-gamedev/tree/zstbtt) in case anybody wants to pick it up.

I'm closing this PR but I think it would be cool to have a lib with only the zig port in that we can hack on. What do you guys think?

PR #183, opened 2023-01-07
Make upload_heap_capacity a build option

gmodarelli:
Hello there 👋 

I was recently working on a project where the 24MB upload heap size was not enough.
I thought it would be a good idea to make it configurable. In this PR I'm proposing the addition of a new build option so that users of the library can choose what best fits their renderer.

Another alternative could be to have this and other values configurable at init time instead, but that seemed like a bigger change.

I'm open to feedback of course

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -18,6 +18,11 @@ pub fn build(b: *std.build.Builder) void {
             "zd3d12-enable-gbv",
             "Enable DirectX 12 GPU-Based Validation (GBV)",
         ) orelse false,
+        .zd3d12_upload_heap_capacity = b.option(
+            u32,
+            "zd3d12-upload-heap-capacity",
+            "DirectX 12 Upload Heap Capacity",
+        ) orelse 24 * 1024 * 1024,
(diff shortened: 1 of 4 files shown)


michal-z:
Thanks!

I think that build option is a good solution but I'm not sure if we need a command line option. I think that heap capacity should be fixed per application, so in `build.zig`:

```zig
    const zd3d12_options = zd3d12.BuildOptionsStep.init(b, .{
        .enable_debug_layer = options.zd3d12_enable_debug_layer,
        .enable_gbv = options.zd3d12_enable_gbv,
        .upload_heap_capacity = 32 * 1024 * 1024, // custom heap capacity for my application
        .enable_d2d = true,
    });
```

Would that work for you or you need to set the capacity from command line?

gmodarelli:
Hey, no, you're right. I don't need a command line option :) I'll update the PR

michal-z:
Perfect!

Btw, great to see you back :)

gmodarelli:
Thanks, I'm glad to be back :)

PR #184, opened 2023-01-08
feat(zstbi): add image resizing

yusdacra:
This adds image resizing using the `stbi_image_resize` library to the `zstbi` package. I'm not sure if the `STBIR_MALLOC` and `STBIR_FREE` implementations are correct since I have no idea what `context` is supposed to be so I just delegated to `zstbiMalloc` and `zstbiFree` respectively. It seems to work fine in my application however, with no issues. I'm not sure either if this is the API wanted for something like this, so it's fine if it's not accepted in this state, just opening so something can get in `zstbi` since resizing is a quite common operation in my experience...


michal-z:
Thanks for the PR!

Looks fine, we can safely ignore `context` in alloc functions.

PR #185, opened 2023-01-09
feat(zstbi): implement writing images as jpeg or png files

yusdacra:
Implements writing `zstbi.Image`s as PNG or JPEG to a file using `stb_image_write`. I did not add the other formats `stb_image_write` supports because I do not need those.

### DIFF
--- a/libs/zstbi/src/zstbi.zig
+++ b/libs/zstbi/src/zstbi.zig
@@ -7,11 +7,17 @@ pub fn init(allocator: std.mem.Allocator) void {
     mem_allocator = allocator;
     mem_allocations = std.AutoHashMap(usize, usize).init(allocator);
 
+    // stb image
     zstbiMallocPtr = zstbiMalloc;
     zstbiReallocPtr = zstbiRealloc;
     zstbiFreePtr = zstbiFree;
+    // stb image resize
     zstbirMallocPtr = zstbirMalloc;
     zstbirFreePtr = zstbirFree;
+    // stb image write
+    zstbiwMallocPtr = zstbiMalloc;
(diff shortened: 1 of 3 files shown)


libs/zstbi/src/zstbi.zig
```diff
@@ -22,6 +28,19 @@ pub fn deinit() void {
     mem_allocator = null;
 }
 
+pub const JpegWriteSettings = struct {
+    quality: u32,
+};
+
+pub const ImageWriteFormat = union(enum) {
+    Png,
```
michal-z, line 36:
```suggestion
    png,
```

libs/zstbi/src/zstbi.zig
```diff
@@ -22,6 +28,19 @@ pub fn deinit() void {
...
+    Jpeg: JpegWriteSettings,
```
michal-z, line 37:
```suggestion
    jpg: JpegWriteSettings,
```

libs/zstbi/src/zstbi.zig
```diff
@@ -195,6 +214,27 @@ pub const Image = struct {
         };
     }
 
+    pub fn write_to_file(self: *const Image, filename: [:0]const u8, image_format: ImageWriteFormat) ImageWriteError!void {
```
michal-z, line 217:
```suggestion
    pub fn writeToFile(self: *const Image, filename: [:0]const u8, image_format: ImageWriteFormat) ImageWriteError!void {
```

michal-z (changes requested):
Nice, thanks!

I have added few minor comments (naming issues).
Also, let's use 'jpg` instead 'jpeg' everywhere (it is shorter and easier to write).

PR #186, opened 2023-01-10
zgui: minor fixes and AddInputCharactersUTF8

Avokadoen:
- Added AddInputCharactersUTF8 
- Fix typo in Key enum 
- Convert key modifiers from u32 to Key. This is because the modifiers can't really be registered with addKeyEvent otherwise since zig's type safety does not allow you to cast these values to enum as they are outside enum range.


michal-z:
Thanks!

PR #187, opened 2023-01-10
Add missing IID_IQueryHeap

gmodarelli:
Hello there, this PR simply adds the missing GUID for ID3D12QueryHeap.
I've tested this code on a project I'm working on and it works fine, but let me know if I should add a test or something :)

image


michal-z:
Thanks!

PR #190, opened 2023-01-17
zmath: more decompose utils

Jack-Ji:


michal-z:
Very nice addition, thanks!

PR #193, opened 2023-01-25
samples: add minimal repro of transparency issue

foxnne:
Hi!

You don't need to merge this PR, I just wasn't sure of the easiest way to showcase this. I'm still struggling with the issue of multiple transparent sprites becoming more opaque even when they aren't overlapping.

You should just be able to run using `zig build transparent_quads_wgpu-run`.

I just simply copied the `textured_quad_wgpu` sample and added a second quad, and set the blend states. There is a checkbox in the gui to show the second quad, and you can observe the first quad becoming more opaque when the second quad draws, almost as if the first quad is being redrawn over itself.

I've been wracking my brain over this for quite a while, and I've tried every combination of blend state that I can think of, and I'm beginning to think it doesn't have anything to do with the blend state but something going on with wgpu. I thought maybe trying to reproduce this as minimally as possible would reveal something but the behavior seems to be the same. 

Anyway, if you have time to take a look at this it would be much appreciated, Thanks!

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -47,6 +47,7 @@ pub fn build(b: *std.build.Builder) void {
     installDemo(b, procedural_mesh_wgpu.build(b, options), "procedural_mesh_wgpu");
     installDemo(b, triangle_wgpu.build(b, options), "triangle_wgpu");
     installDemo(b, textured_quad_wgpu.build(b, options), "textured_quad_wgpu");
+    installDemo(b, transparent_quads_wgpu.build(b, options), "transparent_quads_wgpu");
     installDemo(b, gui_test_wgpu.build(b, options), "gui_test_wgpu");
     installDemo(b, audio_experiments_wgpu.build(b, options), "audio_experiments_wgpu");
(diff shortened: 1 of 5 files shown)


michal-z:
Hi!

When drawing 2 quads you call `drawIndexed()` with `index_count = 12` and `instance_count = 2` this is not correct (if my understanding is fine).

`instance_count` is a parameter telling how many times you want to draw *the same* geometry.

So in this case it will draw 2 quads 2 times (it will transfer 12 indices two times).

`drawIndexed(12, 1, 0, 0, 0)` - draws two quads (indices 0 - 5 first quad, indices 6 - 11 second quad)
`drawIndexed(12, 2, 0, 0, 0)` - will basically draw two times above geometry (so it will draw 2 quads 2 times)

Please let me know if it helps to solve the problem.

foxnne:
Thank you so much for explaining, I figured it was something I was misunderstanding, I apologize for any time wasted! Thanks again!

michal-z:
No problem at all! I'm glad I could help.

PR #195, opened 2023-01-27
d3d12: Add support for IDXGIAdapter3

gmodarelli:
Hello there, this PR adds support for IDXGIAdapter3. I need this to query info about the memory usage of my apps via `IDXGIAdapter3.QueryVideoMemoryInfo`.

I've tested that everything builds correctly and I've run the `bindless` sample with Debug Layers on and everything works correctly without any additional warnings or errors.

I've also exposed the `adapter` via the `GraphicsContext` and I release it when the `GraphicsContext` gets deinitialized. I'm not sure if there's another way to get the `adapter` from the `device`. If there's a way I'm okay with removing the `adapter` from `GraphicsContext`.

Cheers!

### DIFF
--- a/libs/zd3d12/src/zd3d12.zig
+++ b/libs/zd3d12/src/zd3d12.zig
@@ -50,6 +50,7 @@ pub const GraphicsContext = struct {
     const max_num_buffered_resource_barriers = 16;
 
     device: *d3d12.IDevice9,
+    adapter: *dxgi.IAdapter3,
     cmdqueue: *d3d12.ICommandQueue,
     cmdlist: *d3d12.IGraphicsCommandList6,
     cmdallocs: [max_num_buffered_frames]*d3d12.ICommandAllocator,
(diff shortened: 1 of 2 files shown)


gmodarelli:
I found an error in the new functions I've exposed via the `IDXGIAdapter3`. I'm fixing them now, sorry.

gmodarelli:
Okay, I've fixed it now and tested it in my application. It works correctly 😊 
image
(Values are shown in bytes)

gmodarelli:
And finally this is how I query the memory usage at the end of the frame.
image

I've found a [GetAdapter](https://learn.microsoft.com/en-us/windows/win32/api/dxgi/nf-dxgi-idxgidevice-getadapter) function I could use to get the `adapter` and store it in my own graphics state. So if you prefer it, I can remove the adapter from `GraphicsContext`

michal-z:
Very nice, thanks! I like having an `adapter` in `GraphicsContext`.

PR #196, opened 2023-01-28
Add support for loading 64bpp RGBA textures

gmodarelli:
Hello again, I've encountered some 64bbp RGBA textures on [Poly Haven](https://polyhaven.com/) and had to add support for them to wincoded.zig and zd3d12.zig.

I also have a question a bit related to this. Have you planned or thought about adding support for loading DDS textures? I used them in my C++ experiments via [DirectXTK12](https://github.com/Microsoft/DirectXTK12) and they're really nice to work with. It might also be something I can help with if you want.

Cheers

### DIFF
--- a/libs/zd3d12/src/zd3d12.zig
+++ b/libs/zd3d12/src/zd3d12.zig
@@ -1463,6 +1463,7 @@ pub const GraphicsContext = struct {
             if (eql(u8, asBytes(&pixel_format), asBytes(&wic.GUID_PixelFormat32bppBGR))) break :blk 4;
             if (eql(u8, asBytes(&pixel_format), asBytes(&wic.GUID_PixelFormat32bppBGRA))) break :blk 4;
             if (eql(u8, asBytes(&pixel_format), asBytes(&wic.GUID_PixelFormat32bppPBGRA))) break :blk 4;
+            if (eql(u8, asBytes(&pixel_format), asBytes(&wic.GUID_PixelFormat64bppRGBA))) break :blk 4;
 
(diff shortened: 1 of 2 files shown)


michal-z:
Thanks! Yes I would like to support DDS format, looks like WIC (wincodec) has support for it: https://learn.microsoft.com/en-us/windows/win32/wic/dds-format-overview

gmodarelli:
Nice, I'll add support for them then 🥳

gmodarelli:
I've done some tests, but it looks like WIC only supports BC1, BC2 and BC3 encoded images. It doesn't support any of the DXGI formats plus all other BCn formats.

https://learn.microsoft.com/en-us/windows/win32/wic/dds-format-overview#pixel-format-support

michal-z:
I see. How is DIrectX TK decoding those textures? Maybe we can borrow/adapt their implementation?

gmodarelli:
They wrote a custom DDS reader, so I was thinking that we could do something similar to how you integrated mesh_optimizer for example. But I'm not sure how easy it is to integrate give it's a C++ API.

https://github.com/microsoft/DirectXTK12/blob/main/Inc/DDSTextureLoader.h
https://github.com/microsoft/DirectXTK12/blob/main/Src/DDSTextureLoader.cpp

Some of the included headers won't be needed as they are dealing with uploading data to the GPU via D3D12, so not really DDS-specific.

gmodarelli:
I've started implementing this. I'm at the very start but I believe it should be achievable.
I'm making a quick & dirty implementation so I can test that the textures are loaded correctly from my app and then I'll implement proper error reporting and will try to integrate it in `zd3d12`
image

michal-z:
Awesome!

PR #199, opened 2023-02-01
zmath: Add `perspectiveFovLhGl()`

castholm:
I thought it might be useful to provide a left-handed variant of `perspectiveFovRhGl()` that also makes use of the entire OpenGL default [-1, 1] range and yields the desired result. This neatly completes the set of the four `perspectiveFov()` variants.

The previous perspective functions didn't have any tests and I wouldn't know where to start 😅 but if a demonstration is enough, here are the results of using the four `perspectiveFov()` to render a clipped cube in OpenGL. The cube is 2x2x2 and positioned in the center of a frustum where *far* - *near* = 1. Only the GL variants clip correctly.

image


michal-z:
Thanks for the PR and for the demonstration!

PR #201, opened 2023-02-01
zmath: Fix orthographic functions; add Z [-1, 1] variants

castholm:
- Both `orthographicOffCenter()` functions were mistakenly defined as column-major and didn't work, so I transposed them to row-major.
- Added OpenGL variants of all orthographic functions that work with Z [-1, 1].

I've tested and verified all 8 orthographic functions locally and they seem to behave as expected. Let me know if you want me to provide anything.


michal-z:
Thanks!

PR #210, opened 2023-02-12
zbullet: Add Body.setCollisionFlags

chip2n:
Needed these to disable collision for "sensor-like" objects. Lots of other neat things you can do with them apparently, like disabling debug drawing for a body.

### DIFF
--- a/libs/zbullet/libs/cbullet/cbullet.cpp
+++ b/libs/zbullet/libs/cbullet/cbullet.cpp
@@ -1520,6 +1520,12 @@ void cbtBodySetCcdMotionThreshold(CbtBodyHandle body_handle, float threshold) {
     body->setCcdMotionThreshold(threshold);
 }
 
+void cbtBodySetCollisionFlags(CbtBodyHandle body_handle, int flags) {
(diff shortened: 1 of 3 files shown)


michal-z:
Thanks for the PR!

Btw, is your project open source?

chip2n:
It's not open source yet, but probably will be once it's a bit more presentable. I'll let you know when that happens! It's a little mini golf game :)

You've done great work on this collection of libraries by the way. I'm amazed I haven't encountered any build issue on either Linux or Mac.

Also, quick question: there's currently libraries for both Bullet and Jolt physics in this repo. I've only ever used Bullet; is Jolt worth switching to? I guess I don't need fancy features for this game, but I'm curious about the advantages of Jolt.

michal-z:
Yes, please let me know once you open source it!

Thanks, I'm glad you like it.

Generally, Jolt Physics is much faster and scales much better if the number of CPU cores. Take a look at this document: https://jrouwe.nl/jolt/JoltPhysicsMulticoreScaling.pdf

Also, Jolt is actively maintained but Bullet not so much.

PR #214, opened 2023-02-17
zstbi: Add `writeToFn` and externs

foxnne:
Added `writeToFn` and externs to handle both png and jpg cases similar to `writeToFile`. 

These are useful for cases such as mine where you may want to use a context to write images into a zipped file. 

I do have a question though that could result in adding to this PR:

How do you create a new blank `Image` properly? I'm currently just trying to create the struct as any other struct and allocating bytes for an image width and height, but then I get an error in other code from `Image.deinit()`, as well as the resulting .png seems to be full grey and not transparent as I would have suspected with zeroed data. I suspect it's because there isn't bytes in `data` for the signature and data blocks, that `init` would include. I also tried passing my data to `initFromData`, which works fine when reading from a file, but with just allocated color data, it errors.

### DIFF
--- a/libs/zstbi/src/zstbi.zig
+++ b/libs/zstbi/src/zstbi.zig
@@ -239,6 +239,33 @@ pub const Image = struct {
         }
     }
 
+    pub fn writeToFn(
+        self: *const Image,
+        write_fn: *const fn (?*anyopaque, ?*anyopaque, c_int) callconv(.C) void,
+        context: ?*anyopaque,
+        image_format: ImageWriteFormat,
+    ) ImageWriteError!void {
+        const w = @intCast(c_int, self.width);
+        const h = @intCast(c_int, self.height);
+        const comp = @intCast(c_int, self.num_components);
+        const result = switch (image_format) {
+            .png => stbi_write_png_to_func(write_fn, context, w, h, comp, self.data.ptr, 0),
+            .jpg => |settings| stbi_write_jpg_to_func(
+                write_fn,
+                context,
+                w,
+                h,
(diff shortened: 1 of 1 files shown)


michal-z:
Thanks!

You should be able to create a new blank image just by creating an instance of `zstbi.Image` struct. You need to correctly set all the fields and allocate the memory yourself using the same allocator which was passed to `zstbi.init()`.

foxnne:
Hmmm... I can't seem to figure out what's going on then. Below is the code I'm using:
```
...
    const layer_texture = pixi.gfx.Texture.init(pixi.state.gctx, internal.width, internal.height, .{});
    const layer_image_data = try pixi.state.allocator.alloc(u8, internal.width * internal.height * 4);
    var layer: pixi.storage.Internal.Layer = .{
        .name = try std.fmt.allocPrintZ(pixi.state.allocator, "{s}", .{"Layer 0"}),
        .texture_handle = layer_texture.handle,
        .texture_view_handle = layer_texture.view_handle,
        .data = layer_image_data,
        .image = pixi.gfx.createImage(layer_image_data, internal.width, internal.height),
    };

    try internal.layers.append(layer);
...
```

The only thing I can think is that I'm calling allocate to a variable that only belongs to the function scope? So that data.ptr becomes null when the function is done being called? I'm trying to figure out a good way to avoid that. Though I do this pattern elsewhere and don't seem to have an issue...

michal-z:
Looks like `layer_image_data` is stored twice, in `layer.data` and in `layer.image.data`. Double free, maybe?

PR #215, opened 2023-02-17
zgui: Expose `beginPopupContextItem`

foxnne:
This is useful for adding context menus to tree nodes items and similar.


michal-z:
Thanks! Implementation in `zgui.cpp` will be also needed.

foxnne:
Should be good to go now.

PR #216, opened 2023-02-19
Update to new for loop syntax

kcbanner:
This updates the new for loop syntax to support latest zig master. 

I also included a small change to the zmath build script to fix running the tests.


michal-z:
Nice! Thanks for the PR. I will merge it as soon as Zig official binaries hit ziglang.org/download

PR #218, opened 2023-02-23
[zsdl] bindings wip

hazeycode:
added a few bindings/wrappers

### DIFF
--- a/libs/zsdl/src/zsdl.zig
+++ b/libs/zsdl/src/zsdl.zig
@@ -123,6 +123,13 @@ pub const Window = opaque {
     /// `pub fn destroy(window: *Window) void`
     pub const destroy = SDL_DestroyWindow;
     extern fn SDL_DestroyWindow(window: *Window) void;
+
+    pub fn getDisplayMode(window: *Window) DisplayMode {
+        var mode: DisplayMode = undefined;
+        SDL_GetWindowDisplayMode(window, &mode);
+        return mode;
+    }
+    extern fn SDL_GetWindowDisplayMode(*Window, *DisplayMode) void;
 };
 
 pub const EventType = enum(u32) {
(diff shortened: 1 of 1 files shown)


libs/zsdl/src/zsdl.zig
```diff
@@ -899,6 +906,45 @@ pub fn setHint(name: [:0]const u8, value: [:0]const u8) bool {
 }
 extern fn SDL_SetHint(name: [*:0]const u8, value: [*:0]const u8) i32;
 
+pub const SDL_DisplayID = u32;
```
michal-z, line 909:
```suggestion
pub const DisplayId = u32;
```

libs/zsdl/src/zsdl.zig
```diff
@@ -899,6 +906,45 @@ pub fn setHint(name: [:0]const u8, value: [:0]const u8) bool {
...
+
+pub const DisplayMode = DisplayMode_SDL2;
+
+pub const DisplayMode_SDL2 = extern struct {
```
michal-z, line 913:
```suggestion
const DisplayMode_SDL2 = extern struct {
```
This could be private for now.

libs/zsdl/src/zsdl.zig
```diff
@@ -899,6 +906,45 @@ pub fn setHint(name: [:0]const u8, value: [:0]const u8) bool {
...
+    format: u32,
+    w: c_int,
+    h: c_int,
+    refresh_rate: c_int,
+    driverdata: *anyopaque,
+};
+
+pub const DisplayMode_SDL3 = extern struct {
```
michal-z, line 921:
```suggestion
const DisplayMode_SDL3 = extern struct {
```
This could be private for now.

libs/zsdl/src/zsdl.zig
```diff
@@ -899,6 +906,45 @@ pub fn setHint(name: [:0]const u8, value: [:0]const u8) bool {
...
+    displayID: SDL_DisplayID,
```
michal-z, line 922:
```suggestion
    display_id: DisplayId,
```

michal-z (changes requested):
Thanks! Few naming things.

PR #219, opened 2023-02-24
zmesh: expose zcgltf package

### LINKED ISSUE #217: zmesh: zcgltf types are not available to user
Hi michal, I'm trying to load the whole scene defined by GLTF file.
Just found that zcgltf types are not available to user at all, which make it
impossible to write recursive function to load a tree of nodes.
Can you make them public somehow?

Jack-Ji:
fix #217 

Not sure if this is good solution, just a quick fix.


michal-z:
I think this is fine. Thanks!

PR #220, opened 2023-02-24
[zsdl] zigify pollEvent

hazeycode:
Should we prefer zigified APIs like this?

### DIFF
--- a/libs/zsdl/src/zsdl.zig
+++ b/libs/zsdl/src/zsdl.zig
@@ -894,8 +894,9 @@ pub const Event = extern union {
     }
 };
 
-pub fn pollEvent(event: ?*Event) bool {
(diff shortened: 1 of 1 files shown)


michal-z:
Yes, this would be much nicer. We would lost one use case though. You can pass `null` to `pollEvent()` to check if there is any event available. From the docs:
```
 * If `event` is NULL, it simply returns 1 if there is an event in the queue,
 * but will not remove it from the queue.
```

hazeycode:
In this case, maybe making an exception for `pollEvent` and keeping it close to C version is best? But zigify where no API surface is lost?

michal-z:
Yes, this is basically what I would like to do. Zigify where possible (without losing functionality).

PR #223, opened 2023-02-24
DDS Textures support

gmodarelli:
Hello, this PR adds support for loading DDS textures. The dds_loader.zig file is a port of the DDS Texture Loader from [Microsoft's DirectXTK12](https://github.com/microsoft/DirectXTK12).

`zd3d12` now exposes a new function to create and upload DDS textures from file: `GraphicsContext.createAndUploadTex2dFromDdsFile()`.

The `dds_loader.zig` modules supports all texture formats and dimensions (1D, 2D, 3D and cubemaps), but those I haven't created loading code for in `zd3d12`.

I've also left out XBox-specific formats because I'd have no way to test them.

I think the PR is good to go, but I'm not sure about my error handling in `dds_loader.zig` and `GraphicsContext.createAndUploadTex2dFromDdsFile()`. Any feedback on this would be appreciated 🙏


libs/zwin32/src/dds_loader.zig
```diff
@@ -0,0 +1,867 @@
...
+    }
+
+    return .{
+        .num_bytes = num_bytes,
+        .row_bytes = row_bytes,
+        .num_rows = num_rows,
+    };
+}
+
+inline fn bitsPerPixel(format: dxgi.FORMAT) usize {
```
michal-z, line 799: We have `dxgi.Format.pixelSizeInBytes()` method, could we use it instead?

libs/zwin32/src/dds_loader.zig
```diff
@@ -0,0 +1,867 @@
...
+
+        .R1_UNORM => 1,
+
+        .BC1_TYPELESS, .BC1_UNORM, .BC1_UNORM_SRGB, .BC4_TYPELESS, .BC4_UNORM, .BC4_SNORM => 4,
+
+        else => 0,
+    };
+}
+
+inline fn isDepthStencil(format: dxgi.FORMAT) bool {
```
michal-z, line 825: This is a handy function, could we move it to `dxgi.Format`?

michal-z (changes requested):
Few minor things, please take a look.

I think that error handling is fine for now.

libs/zwin32/src/dds_loader.zig
gmodarelli: I don't know what to return though for BC4_ formats as they use 4 bits per pixel 🤔

michal-z: Right, I think we should change it to return pixel size in bits. Should be easy because we use this function in one place only.

libs/zwin32/src/dds_loader.zig
```diff
@@ -339,7 +339,7 @@ pub fn loadTextureFromMemory(file_data: []u8, arena: std.mem.Allocator, device:
         return DdsError.InvalidDDSData;
     }
 
-    if (number_of_planes > 1 and isDepthStencil(format)) {
+    if (number_of_planes > 1 and dxgi.FORMAT.isDepthStencil(format)) {
```
michal-z, line 342:
```suggestion
    if (number_of_planes > 1 and format.isDepthStencil()) {
```

libs/zwin32/src/dds_loader.zig
gmodarelli: On it

gmodarelli:
Thanks for the review, I think I've addressed all the feedback 🙇‍♂️

PR #226, opened 2023-02-25
[zsdl] getMouseFocus and getMouseState

hazeycode:


michal-z:
In this case I would go with below. What do you think?

```zig
pub const MouseState = extern struct {
   x: i32,
   y: i32,
   button_mask: u32,
};

pub fn getMouseState() MouseState {
...
}
```

hazeycode:
I was tempted to this but couldn't decide. If you think this is better then I will agree! I will change and test change as soon as I get on computer today

michal-z:
Thanks!

PR #227, opened 2023-02-26
[zgui] Add sdl2_opengl3 backend

hazeycode:
I have only tested this on Linux so far


michal-z:
I think we could add below code to the `zopengl` (for each function needed by imgui GL renderer) and then we would avoid adding their GL loader.

comptime {
    @export("drawArrays", .{ .name = "glDrawArrays", .linkage = .Strong });
}

Problem with their loader is that they depend on the system OpenGL.framework:
`libgl = dlopen("/System/Library/Frameworks/OpenGL.framework/OpenGL", RTLD_LAZY | RTLD_LOCAL);`

I would prefer to avoid that.

What do you think?

hazeycode:
I didn't know you could do this so easily. It looks good, I will try it out!

michal-z:
Yes this is nice, `compiler_rt` does this a lot.

hazeycode:
Ok I made the change to use the custom loader. I had to add a .h for opengl decls. Currently this fails to compile
```
[chris@chris-desktop proto]$ zig build run
error(compilation): clang failed with stderr: /home/chris/git/proto/zig-gamedev/libs/zgui/libs/imgui/backends/imgui_impl_opengl3.cpp:275:5: error: use of undeclared identifier 'glGetIntegerv'
/home/chris/git/proto/zig-gamedev/libs/zgui/libs/imgui/backends/imgui_impl_opengl3.cpp:276:5: error: use of undeclared identifier 'glGetIntegerv'
...
```

hazeycode:
I'm probably doing something really stupid but right now I don't understand why the forward decls arn't working for functions. Anything `#define`'d seems to be fine but any procs declared `GLAPI` (alias for `extern`) are not apparently undeclared when trying to build imgui_impl_opengl3.cpp

michal-z:
Try adding `#define GL_GLEXT_PROTOTYPES` before including `opengl-decls.h`.

hazeycode:
Thanks! That works. Looks like I just need to export some more of the zopengl symbols to get it to link and it should be good

hazeycode:
I have exported all the procs from zopengl. This crashes the compiler....

michal-z:
This started to happen when you have added exports for all functions? Maybe try `.Weak` linkage mode (Zig code mostly uses this AFAIK).

hazeycode:
Yes If I comment out some of the exports then the crash does not happen. However, this reveals our actual problem; any exports fail to link. For example:
```
error: ld.lld: undefined symbol: glGetProgramInfoLog(unsigned int, int, int*, char*)
    note: referenced by imgui_impl_opengl3.cpp:695 (zig-gamedev/libs/zgui/libs/imgui/backends/imgui_impl_opengl3.cpp:695)
    note:               /home/chris/git/proto-zgui/zig-cache/o/aee57b5c4f097b558f2ef59ee8175fce/imgui_impl_opengl3.o:(CheckProgram(unsigned int, char const*)) in archive /home/chris/git/proto-zgui/zig-cache/o/8de1e8eed125b3d5339f1b1771ef7f1a/libzgui.a
```
I think we need to specify weak linkage on the C side, or suppress undefined symbol errors, so that the static lib can be built?

hazeycode:
Rebased on main branch, to take changes that build zgui as shared lib instead of static but the problem is the same

michal-z:
I think you are right. We need to create a static lib from `zopengl` add `zopengl.Package.link()` and then in root `build.zig`:
`zopengl_pkg.link(exe)` for samples that use `zopengl`. Can you please try that?

hazeycode:
Unfortunately, the result is the same. It feels like this should be possible without building zopengl as separate bin. Having the symbols in the exe should be enough right?
I think the problem is local to zgui, i.e. we need some portable way to mark the GLAPI symbols as weak or some compiler magic to tell the linker that the symbols are coming

hazeycode:
Using `__attribute__((weak))` seems to work.

This leaves the other problem, with the compiler crashing when a certain number of @exports are used...

michal-z:
Hm, I don't understand why we need this. All GL decls are there to satisfy the compiler and our exports in `zopengl` should satisfy the linker.

I guess we can use `__attribute__((weak))` because we compile with clang everywhere. Let's go with it for now if all works. We can leave a comment/warning there.

hazeycode:
> Hm, I don't understand why we need this. All GL decls are there to satisfy the compiler and our exports in `zopengl` should satisfy the linker.

I think it's because zgui is built as a standalone library that is later linked. All strong symbols must be satisfied in a standalone library. I don't think it would be an issue if instead everything was built directly into the final exe for example.

michal-z:
Yeah, make sense.

hazeycode:
Had another go at this. Compiler doesn't crash anymore. But I still can't get it to work. Doesn't look like the symbols are getting exported properly.

Maybe we have to write the function by hand? Since the fn ptrs are defined initially to `undefined`.

readelf output snippet:
```
  8254: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _Z12glGetStringijj
  8255: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _Z15glDeleteBuff[...]
  8256: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _Z15glDeleteProgramj
  8257: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _Z14glCreateShaderj
  8258: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _Z14glShaderSour[...]
  8259: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _Z15glCompileShaderj
  8260: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _Z15glCreateProgramv
  8261: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _Z14glAttachShaderjj
  8262: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _Z13glLinkProgramj
  8263: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _Z14glDetachShaderjj
  8264: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _Z14glDeleteShaderj
  8265: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _Z20glGetUniform[...]
  8266: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _Z19glGetAttribL[...]
  8267: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _Z12glGenBuffersiPj
  8268: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _Z13glBindTexturejj
  8269: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _Z12glBindBufferjj
...
```

michal-z:
I see what's going on - `extern "C"` is missing in `opnegl-decls.h`.

hazeycode:
Thanks! The names look better now
```
8215: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND glGetStringi
8216: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND glDeleteBuffers
8217: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND glDeleteProgram
...
```

But I still crash at runtime. Like this
```
Segmentation fault at address 0x0
???:?:?: 0x0 in ??? (???)
/home/chris/git/protoed-zgui/zig-gamedev/libs/zgui/src/backend_sdl2_opengl3.zig:5:32: 0x2a5fd3 in init (proto)
    if (!ImGui_ImplOpenGL3_Init(null)) unreachable;
                               ^
/home/chris/git/protoed-zgui/src/main-sdl.zig:64:22: 0x2a51ae in main (proto)
    zgui.backend.init(window, gl_context);
                     ^
/home/chris/zig/0.11.0-dev.1817+f6c934677/files/lib/std/start.zig:617:37: 0x2a6acc in main (proto)
            const result = root.main() catch |err| {
                                    ^
```

Of which I am skeptical, because `ImGui_ImplOpenGL3_Init` looks fine
```
8080: 00000000004dc820   731 FUNC    GLOBAL DEFAULT   15 ImGui_ImplOpenGL3_Init
```

michal-z:
Did you call `zopengl.loadCoreProfile()` before initializing `imgui` renderer?

hazeycode:
Yes, like this
```
    try gl.loadCoreProfile(sdl.gl.getProcAddress, gl_major, gl_minor);

    zgui.init(root_allocator);
    zgui.backend.init(window, gl_context);
```

michal-z:
I think that the crash is somewhere in `ImGui_ImplSDL2_InitForOpenGL()` but callstack does not show it. Maybe we should now switch to `strong` linkage?

hazeycode:
I just double checked that the code works without our custom loader.

Also Strong linkage does not help, the result looks the same as with Weak
```
  8215: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND glGetStringi
  8216: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND glDeleteBuffers
  8217: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND glDeleteProgram
...
```

michal-z:
Did you try to remove the weak attrib you added in `opengl-decls.h`?

hazeycode:
I didn't. That brings us back to the linker errors
```
error: ld.lld: undefined symbol: glGetProgramInfoLog
    note: referenced by imgui_impl_opengl3.cpp:697 (zig-gamedev/libs/zgui/libs/imgui/backends/imgui_impl_opengl3.cpp:697)
```

hazeycode:
It crashes on the first gl fn call

```
(lldb) n
Process 1137285 stopped
* thread #1, name = 'proto', stop reason = step over
    frame #0: 0x00000000004dc906 proto`::ImGui_ImplOpenGL3_Init(glsl_version=0x0000000000000000) at imgui_impl_opengl3.cpp:278:5
   275 	#if !defined(IMGUI_IMPL_OPENGL_ES2)
   276 	    GLint major = 0;
   277 	    GLint minor = 0;
-> 278 	    glGetIntegerv(GL_MAJOR_VERSION, &major);
   279 	    glGetIntegerv(GL_MINOR_VERSION, &minor);
   280 	    if (major == 0 && minor == 0)
   281 	    {
(lldb) s
Process 1137285 stopped
* thread #1, name = 'proto', stop reason = signal SIGSEGV: invalid address (fault address: 0x0)
    frame #0: 0x0000000000000000
```

michal-z:
This is what I thought, thanks for checking. I have one more idea. Let's remove the line:
```
#define GL_GLEXT_PROTOTYPE
```

And then, let's declare (in `imgui_impl_opengl3.cpp`) all the function needed as:
```
extern PFNGLGETINTEGERVPROC glGetIntegerv;
...
```

hazeycode:
I hacked in `glGetIntegerv` this way but it result is the same unfortunately

michal-z:
Do you set the `export_c_api = .core` option?

hazeycode:
Yes, this is my build.zig https://zigbin.io/8a898a

hazeycode:
I will try building zopengl as a separate static lib, as you said before

michal-z:
We are a bit closer :)

michal-z:
Maybe now we could remove the weak attrib from `opengl-decls.h`?

michal-z:
The other way we could solve it is to pass our OpenGL loader to `imgui` renderer:

```zig
    ...
    
    try gl.loadCoreProfile(sdl.gl.getProcAddress, gl_major, gl_minor);

    zgui.init(root_allocator);
    zgui.backend.init(window, gl_context, sdl.gl.getProcAddress);
```

In `imgui_impl_opengl3.cpp`:

extern PFNGLGETINTEGERVPROC glGetIntegerv;

bool ImGui_ImplOpenGL3_Init(...)
{
   glGetIntegerv = load("glGetIntegerv");
   ...
}
```

hazeycode:
I think we need the weak attrib to build zgui still. But if I remove it, another problem is revealed:

imgui_impl_opengl3.cpp uses `glBlendFuncSeparate` which is gl core 4.0!

michal-z:
I think we should pass the loader function and load GL functions ourselves - we will have to "approximate" `glBlendFuncSeparate()` with `glBlendFunc()`.

hazeycode:
It feels like it should be possible to simply build zgui with weak symbols that are provided by an exe linking it. Maybe there is some way to get Zig to do this? Or it is a bug?

But yeah, passing the loader function over seems like the way to go!

michal-z:
> It feels like it should be possible to simply build zgui with weak symbols that are provided by an exe linking it. Maybe there is some way to get Zig to do this? Or it is a bug?

Yes, I agree, it should work, imo.

kanewallmann:
The problem is that zopengl is exporting those symbols as function pointers but "opengl-decl.h" and GL_GLEXT_PROTOTYPES declares them as just external functions.

So when the functions are called, it's segfaulting because it's trying to jump execution to the memory location of the function pointer instead of dereferencing the pointer and jumping there.

I fixed it with these changes, but I'm not sure if it's the "right" way: https://github.com/kanewallmann/zig-gamedev/commit/074dcf6f4e8914ca2bf457d0fab62f292f649b80

hazeycode:
Closing because this is very out of date. This should be easier now that https://github.com/michal-z/zig-gamedev/pull/417 was merged

PR #229, opened 2023-02-26
zpool: Make handles extern for better C interop

chip2n:
I found myself wanting to return handles from exported functions for FFI use. I think it makes sense to make handles `extern` to make this easier, since handles are just a wrapped integer anyways.

(There shouldn't be any performance penalty doing this, right?)


michal-z:
Yes, this is a very good idea. There won't be any additional performance cost. Thanks!

PR #231, opened 2023-02-28
zstbi: Added HDR support to Image.loadFromMemory()

soenkekrusche:
I am fairly new to zig, but used stbi in the past. I am not very deep into the internas of it.
I tried to puzzle together what has to be done with existing code and the stbi.h

I am not sure if I built zstbi successfully as it is a library. I would appreciate if someone could tell me a way to test if my changes do remotely what I think :)

### DIFF
--- a/libs/zstbi/src/zstbi.zig
+++ b/libs/zstbi/src/zstbi.zig
@@ -146,14 +146,43 @@ pub const Image = struct {
     }
 
     pub fn loadFromMemory(data: []const u8, forced_num_components: u32) !Image {
-        // TODO: Add support for HDR images (https://github.com/michal-z/zig-gamedev/issues/155).
         var width: u32 = 0;
         var height: u32 = 0;
         var num_components: u32 = 0;
         var bytes_per_component: u32 = 0;
         var bytes_per_row: u32 = 0;
+        var is_hdr = false;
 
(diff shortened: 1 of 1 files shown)


libs/zstbi/src/zstbi.zig
```diff
@@ -146,14 +146,43 @@ pub const Image = struct {
...
+                &y,
+                &ch,
+                @intCast(c_int, forced_num_components),
+            );
+            if (ptr == null) return error.ImageInitFailed;
+
+            num_components = if (forced_num_components == 0) @intCast(u32, ch) else forced_num_components;
+            width = @intCast(u32, x);
+            height = @intCast(u32, y);
+            bytes_per_component = 1;
```
michal-z, line 173: We use `f16` for each component so `bytes_per_component` should be 2.

michal-z (changes requested):

michal-z:
Thanks for the PR! It looks good. I found one issue/typo.

michal-z:
To test it we would have to create a small test app (or modify existing sample app). This is not strictly necessary as part of this PR.

soenkekrusche:
I will resolve the issue after work and try to write a test or simple test app, no promises for that tho

PR #232, opened 2023-02-28
[zsdl] fixup SDL_GetWindowDisplayMode return type

hazeycode:
Doesn't actually seem to matter. But better safe than sorry!


michal-z:
Good catch!

PR #236, opened 2023-03-04
zopengl: optional thin "zigified" wrapper (experimental)

hazeycode:
It's nice to have some extra type safety when working with OpenGL to avoid easy mistakes that would otherwise require some kind of runtime checks or debugging to find.

This is a tentative attempt at implementing a very minimal, thin wrapper around the bindings to add some safety and convenience. I am trying to achieve zero or near-zero overhead in release modes and not restrict use of the API in any way.

I've covered a subset of the API here to use as a point of discussion (some require more attention than others):
`texParameteri`, `texImage2D`, `clear`, `clearColor`, `getError`, `drawArrays`, `bindTexture`, `genTextures`, `bindBuffer`, `genBuffers`, `bufferData`, `bufferSubData`, `attachShader`, `compileShader`, `createProgram`, `createShader`, `deleteShader`, `enableVertexAttribArray`, `getAttribLocation`, `getProgramiv`, `getProgramInfoLog`, `getShaderiv`, `getShaderInfoLog`, `getUniformLocation`, `linkProgram`, `shaderSource`, `useProgram`, `uniform1i`, `vertexAttribPointer`

This API subset is used in [this gist](https://gist.github.com/hazeycode/037e0d4c6b2228038bc48cebc3a7fc8f), comparing a simple example renderer using the raw bindings and using the wrapper. There is a diff as well as source code for each version included.

Note: In this revision, the wrapper is exposed at build time using an option:
```zig
zopengl_pkg = zopengl.Package.build(b, .{
    .options = .{ .api = .wrapper }, // .raw to not use the wrapper
});
```

For development and testing purposes, `zopengl.bindings` provides an escape hatch to the raw bindings if needed.

What do you think?

### DIFF
--- a/libs/zopengl/src/wrapper.zig
+++ b/libs/zopengl/src/wrapper.zig
@@ -1219,7 +1219,7 @@ pub fn bufferData(
     assert(size > 0);
     if (builtin.mode == .Debug) {
         if (maybe_bytes) |bytes| {
-            assert(bytes.len >= size);
+            assert(bytes.len == size);
         }
     }
     bindings.bufferData(
--- a/libs/zopengl/src/wrapper.zig
+++ b/libs/zopengl/src/wrapper.zig
@@ -11,8 +11,8 @@ pub const Program = extern struct { name: Uint = 0 };
 pub const Texture = extern struct { name: Uint = 0 };
 pub const Buffer = extern struct { name: Uint = 0 };
 
-pub const UniformLocation = extern struct { location: Uint = 0 };
-pub const VertexAttribLocation = extern struct { location: Uint = 0 };
+pub const UniformLocation = extern struct { location: Uint };
+pub const VertexAttribLocation = extern struct { location: Uint };
 
 pub const Error = enum(Enum) {
     //----------------------------------------------------------------------------------------------
--- a/libs/zopengl/src/wrapper.zig
+++ b/libs/zopengl/src/wrapper.zig
@@ -771,8 +771,8 @@ pub const RGBA16 = bindings.RGBA16;
 pub const VERTEX_ARRAY = bindings.VERTEX_ARRAY;
 
 // pub var drawArrays: *const fn (mode: Enum, first: Int, count: Sizei) callconv(.C) void = undefined;
-pub fn drawArrays(prim_type: PrimitiveType, first: u16, count: u16) void {
-    bindings.drawArrays(@enumToInt(prim_type), first, count);
+pub fn drawArrays(prim_type: PrimitiveType, first: u32, count: u32) void {
+    bindings.drawArrays(@enumToInt(prim_type), @bitCast(Int, first), @bitCast(Sizei, count));
 }
 
 // pub var drawElements: *const fn (
@@ -1180,7 +1180,6 @@ pub const SRC1_ALPHA = bindings.SRC1_ALPHA;
 
 // pub var bindBuffer: *const fn (target: Enum, buffer: Uint) callconv(.C) void = undefined;
 pub fn bindBuffer(target: BufferTarget, buffer: Buffer) void {
-    assert(@bitCast(Uint, buffer) > 0);
     bindings.bindBuffer(@enumToInt(target), @bitCast(Uint, buffer));
 }
 
@@ -1205,19 +1204,14 @@ pub fn genBuffers(buffers: []Buffer) void {
 pub fn bufferData(
     target: BufferTarget,
     size: usize,
-    maybe_bytes: ?[]const u8,
+    bytes: ?[*]const u8,
     usage: BufferUsage,
 ) void {
     assert(size > 0);
-    if (builtin.mode == .Debug) {
-        if (maybe_bytes) |bytes| {
-            assert(bytes.len == size);
-        }
-    }
     bindings.bufferData(
         @enumToInt(target),
         @intCast(Sizeiptr, size),
-        if (maybe_bytes) |bytes| bytes.ptr else null,
+        bytes,
         @enumToInt(usage),
     );
 }
@@ -1530,22 +1524,20 @@ pub fn linkProgram(program: Program) void {
 //     string: [*c]const [*c]const Char,
 //     length: [*c]const Int,
 // ) callconv(.C) void = undefined;
-pub fn shaderSource(shader: Shader, src_ptrs: []const [*]const u8, src_lengths: []const Int) void {
+pub fn shaderSource(shader: Shader, src_ptrs: []const [*:0]const u8, src_lengths: []const u32) void {
     assert(@bitCast(Uint, shader) > 0);
     assert(src_ptrs.len > 0);
     assert(src_ptrs.len == src_lengths.len);
-    assert(src_ptrs.len <= std.math.maxInt(Int));
     bindings.shaderSource(
         @bitCast(Uint, shader),
-        @intCast(Sizei, src_ptrs.len),
+        @bitCast(Sizei, @intCast(u32, src_ptrs.len)),
         @ptrCast([*c]const [*c]const Char, src_ptrs),
-        src_lengths.ptr,
+        @ptrCast([*c]const Int, src_lengths.ptr),
     );
 }
 
 // pub var useProgram: *const fn (program: Uint) callconv(.C) void = undefined;
 pub fn useProgram(program: Program) void {
-    assert(@bitCast(Uint, program) > 0);
     bindings.useProgram(@bitCast(Uint, program));
 }
 
--- a/libs/zopengl/src/wrapper.zig
+++ b/libs/zopengl/src/wrapper.zig
@@ -689,12 +689,15 @@ pub fn clearColor(r: f32, g: f32, b: f32, a: f32) void {
 
 // pub var getError: *const fn () callconv(.C) Enum = undefined;
 pub fn getError() Error {
-    const try_err = std.meta.intToEnum(Error, bindings.getError());
-    if (std.meta.isError(try_err)) {
-        assert(false);
-        return .no_error;
+    const err_int = bindings.getError();
+    inline for (@typeInfo(Error).Enum.fields) |f| {
+        const this_tag_value = @field(Error, f.name);
+        if (err_int == @enumToInt(this_tag_value)) {
+            return this_tag_value;
+        }
     }
-    return try_err catch unreachable;
+    assert(false);
+    return .no_error;
 }
 
 // pub var getFloatv: *const fn (pname: Enum, data: [*c]Float) callconv(.C) void = undefined;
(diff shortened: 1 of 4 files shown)


libs/zopengl/src/wrapper.zig
```diff
@@ -0,0 +1,2637 @@
+const std = @import("std");
+const assert = std.debug.assert;
+
+pub const bindings = @import("bindings.zig");
+
+pub const Framebuffer = packed struct { _: Uint = 0 };
```
michal-z, line 6: Maybe it would be better to use `extern struct` or `packed struct(u32)`?

libs/zopengl/src/wrapper.zig
```diff
@@ -0,0 +1,2637 @@
...
+
+// pub var bindTexture: *const fn (target: Enum, texture: Uint) callconv(.C) void = undefined;
+pub fn bindTexture(target: TextureTarget, texture: Texture) void {
+    bindings.bindTexture(@enumToInt(target), @bitCast(Uint, texture));
+}
+
+// pub var deleteTextures: *const fn (n: Sizei, textures: [*c]const Uint) callconv(.C) void = undefined;
+
+// pub var genTextures: *const fn (n: Sizei, textures: [*c]Uint) callconv(.C) void = undefined;
+pub fn genTextures(ptr_or_slice: anytype) void {
```
michal-z, line 842: This function is a bit complicated. Instead, I would prefer to have a function like below for each resource type (I think it is fine to generate only one resource name at a time):

```zig
pub fn genTexture() Texture {
   var name: Uint = undefined;
   bindings.genTextures(1, &name);
   return .{ ._ = name };
}
```

libs/zopengl/src/wrapper.zig
```diff
@@ -0,0 +1,2637 @@
...
+// pub var bufferData: *const fn (
+//     target: Enum,
+//     size: Sizeiptr,
+//     data: ?*const anyopaque,
+//     usage: Enum,
+// ) callconv(.C) void = undefined;
+pub fn bufferData(
+    target: BufferTarget,
+    size: usize,
+    bytes: ?[*]const u8,
```
michal-z, line 1229: I think we should pass an optional slice here?

libs/zopengl/src/wrapper.zig
```diff
@@ -0,0 +1,2637 @@
...
+// ) callconv(.C) void = undefined;
+pub fn texImage2D(args: struct {
+    target: TextureTarget,
+    level: usize,
+    internal_format: TextureInternalFormat,
+    width: usize,
+    height: usize,
+    format: PixelFormat,
+    pixel_type: PixelType,
+    data: ?[*]const u8,
```
michal-z, line 626: Maybe it should be slice? We could then do an additional check:
```zig
assert(data.size == width * height * format.sizeInBytes());
```

michal-z (changes requested):
This is really nice, thanks!

I've added few comments.

libs/zopengl/src/wrapper.zig
michal-z: Well actually let's leave it as it is for now. Size is specified by `width`, `height` and `pixel_type`. What do you think?

libs/zopengl/src/wrapper.zig
michal-z: I think we should name the member `name` because this is how OpenGL refers to resource ids.
```zig
pub const Buffer = extern struct { name: Uint = 0 };
```

libs/zopengl/src/wrapper.zig
hazeycode: I'm not sure if you'd ever want to "underrun" the texture buffer? I think your suggestion makes sense. However we do have to unwrap the optional if it is a slice, like the updated bufferData: https://github.com/michal-z/zig-gamedev/pull/236/commits/9eb073af46a522f6bceb73729c0dbfb2d54e0162

libs/zopengl/src/wrapper.zig
hazeycode: I added some asserts also. Note that the optional slice has to be unwrapped. What do you think?

libs/zopengl/src/wrapper.zig
hazeycode: Looks good. I was unsure about `packed` vs `extern` here.

In this case, do you think the `@bitCast`s should be changed to `handle.name` or it doesn't matter?

libs/zopengl/src/wrapper.zig
hazeycode: I didn't like it much either. I'd like to be able to keep the multi object generation pattern. What about providing both `genTexture` and `genTextures` as in updated code?

hazeycode:
A current flaw with this is that it introduces `unreachable`s where the user may not like. As an example, `@intToEnum` introduces an implicit `unreachable` if the driver returns some bad value.
```zig
pub fn getError() Error {
    return @intToEnum(Error, bindings.getError());
}
```
In many cases the game could just carry on running instead of crashing or triggering some undefined behavior, depending on the optimization mode. However, just crashing is probably preferred in a development build.

libs/zopengl/src/wrapper.zig
```diff
@@ -0,0 +1,2639 @@
...
+// ) callconv(.C) void = undefined;
+pub fn shaderSource(shader: Shader, src_ptrs: []const [*]const u8, src_lengths: []const usize) void {
+    assert(@bitCast(Uint, shader) > 0);
+    assert(src_ptrs.len > 0);
+    assert(src_ptrs.len == src_lengths.len);
+    bindings.shaderSource(
+        @bitCast(Uint, shader),
+        @intCast(Sizei, src_ptrs.len),
+        @ptrCast([*c]const [*c]const Char, src_ptrs),
+        @ptrCast([*c]const Int, src_lengths),
```
hazeycode, line 1557: This looks like a bug. A slice of `usize`s is passed in but we treat it like a pointer to some `Int`s, which has a different size?
I only tested this with one source ptr and len. Needs to be tested more.

libs/zopengl/src/wrapper.zig
michal-z: Make sense but `genTexture()` should return `Texture` so that a user can declare it `const`:
```zig
const texture = gl.genTexture();
```

libs/zopengl/src/wrapper.zig
michal-z: When using `extern struct` it is guranteed that there won't be any padding (this C rule). We don't need to check the size in this case.

I think that `handle.name` would be cleaner than `@bitCast()` but I don't have strong opinion.

libs/zopengl/src/wrapper.zig
```diff
@@ -0,0 +1,2624 @@
...
+pub const RGBA4 = bindings.RGBA4;
+pub const RGB5_A1 = bindings.RGB5_A1;
+pub const RGBA8 = bindings.RGBA8;
+pub const RGB10_A2 = bindings.RGB10_A2;
+pub const RGBA12 = bindings.RGBA12;
+pub const RGBA16 = bindings.RGBA16;
+pub const VERTEX_ARRAY = bindings.VERTEX_ARRAY;
+
+// pub var drawArrays: *const fn (mode: Enum, first: Int, count: Sizei) callconv(.C) void = undefined;
+pub fn drawArrays(prim_type: PrimitiveType, first: u16, count: u16) void {
```
michal-z, line 774: `first` and `count` should be `u32`.

libs/zopengl/src/wrapper.zig
```diff
@@ -0,0 +1,2624 @@
...
+// pub var isQuery: *const fn (id: Uint) callconv(.C) Boolean = undefined;
+// pub var beginQuery: *const fn (target: Enum, id: Uint) callconv(.C) void = undefined;
+// pub var endQuery: *const fn (target: Enum) callconv(.C) void = undefined;
+// pub var getQueryiv: *const fn (target: Enum, pname: Enum, params: [*c]Int) callconv(.C) void = undefined;
+// pub var getQueryObjectiv: *const fn (id: Uint, pname: Enum, params: [*c]Int) callconv(.C) void = undefined;
+// pub var getQueryObjectuiv: *const fn (id: Uint, pname: Enum, params: [*c]Uint) callconv(.C) void = undefined;
+
+// pub var bindBuffer: *const fn (target: Enum, buffer: Uint) callconv(.C) void = undefined;
+pub fn bindBuffer(target: BufferTarget, buffer: Buffer) void {
+    assert(@bitCast(Uint, buffer) > 0);
```
michal-z, line 1183: It should be possible to bind buffer with `name == 0` to break any existing binding. Some people do this in OpenGL code to make sure that they don't modify bound buffer by mistake.
```zig
gl.bindBuffer(target, buffer);
// Modify buffer content
gl.bindBuffer(target, .{ .name = 0 });
```

libs/zopengl/src/wrapper.zig
```diff
@@ -0,0 +1,2624 @@
...
+    bindings.linkProgram(@bitCast(Uint, program));
+}
+
+// pub var shaderSource: *const fn (
+//     shader: Uint,
+//     count: Sizei,
+//     string: [*c]const [*c]const Char,
+//     length: [*c]const Int,
+// ) callconv(.C) void = undefined;
+pub fn shaderSource(shader: Shader, src_ptrs: []const [*]const u8, src_lengths: []const Int) void {
```
michal-z, line 1533:
```suggestion
pub fn shaderSource(shader: Shader, src_ptrs: []const [*:0]const u8, src_lengths: []const Int) void {
```

libs/zopengl/src/wrapper.zig
```diff
@@ -0,0 +1,2624 @@
...
+        @bitCast(Uint, shader),
+        @intCast(Sizei, src_ptrs.len),
+        @ptrCast([*c]const [*c]const Char, src_ptrs),
+        src_lengths.ptr,
+    );
+}
+
+// pub var useProgram: *const fn (program: Uint) callconv(.C) void = undefined;
+pub fn useProgram(program: Program) void {
+    assert(@bitCast(Uint, program) > 0);
```
michal-z, line 1548: I think it should be possible to use program 0. See comment for `bindBuffer()`.

libs/zopengl/src/wrapper.zig
```diff
@@ -0,0 +1,2624 @@
...
+// pub var bufferData: *const fn (
+//     target: Enum,
+//     size: Sizeiptr,
+//     data: ?*const anyopaque,
+//     usage: Enum,
+// ) callconv(.C) void = undefined;
+pub fn bufferData(
+    target: BufferTarget,
+    size: usize,
+    maybe_bytes: ?[]const u8,
```
michal-z, line 1208: Maybe? This would be consistent with `texImage2D()`?
```suggestion
    maybe_bytes: ?[*]const u8,
```

michal-z (changes requested):
Few minor things and I think we can merge it and start using it.

PR #237, opened 2023-03-05
Expose glfwSetTime in zglfw

bcrist:
Just a tiny change to add a `glfwSetTime` wrapper to zglfw.

It's a good practice when using `glfwGetTime` to occasionally also use `glfwSetTime` to reset the counter (and do whatever else you need to ensure the discontinuity doesn't manifest itself in your simulation logic).  Otherwise you eventually start losing floating point precision and that can cause a lot of subtle issues.


michal-z:
Thanks for the PR!

PR #238, opened 2023-03-05
Expose GLFW functions for windowed mode apps

bcrist:
This exposes the GLFW APIs for:

- Setting window size (getter already exists)
- Getting and setting window position
- Changing window title after creation
- Setting additional window callbacks:
  - Size
  - Framebuffer Size
  - Position
  - Content scale (for when windows move between monitors with different DPI)

Generally most games don't need these since they run in fullscreen, but for tooling and non-game uses, they're a necessity IMHO.


michal-z:
Thanks!

PR #242, opened 2023-03-08
zopengl: more experimental wrappings

hazeycode:
Adds
- `ParameterName` and `StringName` enums
- ~~`getInteger`~~, `getIntegerv` (I'm not sure about this, see below)
- `getString`
- `getStringi`

### DIFF
--- a/libs/zopengl/src/wrapper.zig
+++ b/libs/zopengl/src/wrapper.zig
@@ -30,9 +30,9 @@ pub const Error = enum(Enum) {
 };
 
 pub const Capability = enum(Enum) {
-    //--------------------------------------------------------------------------------------------------
+    //---------------------------------------------------------------------------------------------
     // OpenGL 1.0 (Core Profile)
-    //--------------------------------------------------------------------------------------------------
+    //---------------------------------------------------------------------------------------------
     blend = BLEND,
     cull_face = CULL_FACE,
     depth_test = DEPTH_TEST,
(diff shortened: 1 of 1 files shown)


libs/zopengl/src/wrapper.zig
```diff
@@ -41,40 +41,259 @@ pub const Capability = enum(Enum) {
...
+    draw_buffer13 = DRAW_BUFFER13,
+    draw_buffer14 = DRAW_BUFFER14,
+    draw_buffer15 = DRAW_BUFFER15,
+    draw_framebuffer_binding = DRAW_FRAMEBUFFER_BINDING,
+    fragment_shader_derivative_hint = FRAGMENT_SHADER_DERIVATIVE_HINT,
+    max_combined_texture_image_units = MAX_COMBINED_TEXTURE_IMAGE_UNITS,
+    max_draw_buffers = MAX_DRAW_BUFFERS,
+    max_fragment_uniform_components = MAX_FRAGMENT_UNIFORM_COMPONENTS,
+    max_texture_image_units = MAX_TEXTURE_IMAGE_UNITS,
+    //max_varying_floats = MAX_VARYING_FLOATS, // NOTE: MAX_VARYING_FLOATS is equal to MAX_VARYING_COMPONENTS
```
hazeycode, line 221: urgh!

libs/zopengl/src/wrapper.zig
```diff
@@ -759,7 +978,17 @@ pub fn getError() Error {
...
+
 // pub var getIntegerv: *const fn (pname: Enum, data: [*c]Int) callconv(.C) void = undefined;
+pub fn getInteger(parameter: Parameter) Int {
+    var res: Int = undefined;
+    bindings.getIntegerv(@enumToInt(parameter), &res);
+    return res;
+}
+pub fn getIntegerv(parameter: Parameter, ptr: [*]Int) void {
+    bindings.getIntegerv(@enumToInt(parameter), ptr);
+}
```
hazeycode, line 990: This is awful. Is it really impossible to get some safety here? Is there some way to find out how many bytes opengl is going to from the pointer?

hazeycode: I wanted to see what WebGL does here. And well, it doesn't have these functions. Instead it has a `getParameter` that returns "typed" thing(s) depending on the pname.

Maybe we should try something similar using comptime generics?

hazeycode: I've revised this in https://github.com/michal-z/zig-gamedev/pull/242/commits/ebec59163b96dfc2f3140bcc4f08b313b2de48bb
I think this works well, at least on user's side
````zig
// pub var getIntegerv: *const fn (pname: Enum, data: [*c]Int) callconv(.C) void = undefined;
pub fn getInteger1(pname: ParameterName1) Int {
    var res: Int = undefined;
    bindings.getIntegerv(@enumToInt(pname), &res);
    return res;
}
pub fn getInteger2(pname: ParameterName2) [2]Int {
    var res: [2]Int = undefined;
    bindings.getIntegerv(@enumToInt(pname), &res);
    return res;
}
pub fn getInteger4(pname: ParameterName4) [4]Int {
    var res: [4]Int = undefined;
    bindings.getIntegerv(@enumToInt(pname), &res);
    return res;
}
```
What do you think?

hazeycode: This doesn't work. Some parameters return a variable amount of things. `GL_COMPRESSED_TEXTURE_FORMATS` for example.

libs/zopengl/src/wrapper.zig
michal-z: `MAX_VARYING_FLOATS` is an old name from pre OpenGL 3.0 - we can ignore it.

PR #260, opened 2023-03-25
zpool: Fix broken Pool.init()

chip2n:
Noticed this constructor is not used anywhere, so it was silently broken. Would probably make sense to test adding/removing handles from a 0 capacity pool as well, but for now I just added a simple test so the init() function is compiled and analyzed by Zig.


michal-z:
Nice find, thanks!

PR #262, opened 2023-04-02
ztracy: add macos frameworks search path

hazeycode:
Fixes ztracy build for macOS. Without this, compiling Tracy cpp client fails with `fatal error: 'mach/mach_vm.h' file not found`


michal-z:
Yes, I saw this and didn't know how to fix. Many thanks!

PR #264, opened 2023-04-05
zglfw: Add `setShouldClose`

foxnne:
Useful for apps that want to delay closing the window or handle unsaved documents etc.


michal-z:
Thanks!

PR #278, opened 2023-04-15
Misc zgflw, zmath, and zmesh additions

kcbanner:
These are some local changes I've been using and it's time to PR them! Please let me know if you'd prefer separate PRs per-lib, or if any of these changes aren't desired and I can remove those parts.

zglfw:
- support building as a shared lib
- add VideoMode / getVideoMode
- add WindowHint
- add glfw functions required for OpenGL
- add maybeErrorString

zmath:
- add rotate(quat, vec)
- add quatToPitchYawRoll
- rename quatFromRollPitchYaw -> quatFromPitchYawRoll to match argument order

zmesh:
- support building as a shared lib
- make mem public so library users can construct zcgltf.Options referencing the zmeshAllocUser/zmeshFreeUser


michal-z:
Thanks for the PR!

All looks good but note that `quatFromRollPitchYaw()` name is like this by design.
It expresses the order of rotations (matrix multiplications): `mZ * mX * mY`
The order of args is in natural x, y, z order.
This is the convention borrowed from DirectXMath and I would like to stick to it.

kcbanner:
Ah, that makes sense! I wasn't aware of that convention. Switched it back.

PR #280, opened 2023-04-17
zflecs: Add `set_pair` and `remove_pair`

foxnne:


foxnne:
One thing here: flecs has two functions `set_pair_second` and `set_pair`, which correspond to which component carries the data in the pair. I just followed what you were doing on `add_pair` and `set`, which means `set_pair` and `set_pair_second` would be identical. Currently, the user has to make sure that the pair consists of acceptable types for `set_pair` to work correctly - the first component either must be a tag for the second component to store data, or the first component must not be a tag and that type be used.

PR #283, opened 2023-04-17
Add ImPlot Demo and Shaded Line

ckrowland:
I noticed I like having the ImPlot demo binding handy to look at all the features quickly. It might be nice to have the official binding in zgui.

Also played around a bit with the Shaded line in ImPlot. Thought you might want it as well.


michal-z:
Very nice, thanks!

PR #286, opened 2023-04-22
Add support for loading cubemaps from DDS files

gmodarelli:
Hello there, I don't have a test for this on zig-gamedev, but we're using this code for Image-Based Lighting for Tides of Revival and it seems to work fine :)

### DIFF
--- a/libs/zd3d12/src/zd3d12.zig
+++ b/libs/zd3d12/src/zd3d12.zig
@@ -1545,7 +1545,9 @@ pub const GraphicsContext = struct {
         gctx: *GraphicsContext,
         relpath: []const u8,
         arena: std.mem.Allocator,
-        cubemap: bool,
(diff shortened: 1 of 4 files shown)


michal-z:
Thanks!

I like this pattern for settings like this:

```zig
    pub fn createAndUploadTex2dFromDdsFile(
        gctx: *GraphicsContext,
        relpath: []const u8,
        args: struct {
            is_cubemap: bool = false,
        },
   )
```

This gives us ability to provide a default value and also is easier for the user to read on the call site.

```zig
createAndUploadTex2dFromDdsFile(..., .{ .is_cubemap = true });
```

We use this pattern for `createAndUploadTex2dFromFile()` function.

gmodarelli:
Oh nice, I like it too. Changing it now :)

PR #287, opened 2023-04-26
Fix memory leak in zaudio sample code

Beyley:
Beforehand neither `engine.destroy()` or `music.destroy()` was called, this fixes that, so now this sample code runs and exits cleanly


michal-z:
Thanks!

PR #289, opened 2023-05-06
zpool: Fix incorrect error set in requireLiveHandle

chip2n:
I noticed that `requireLiveHandle` (as well as `getColumns` and probably more) could return `error.PoolIsFull`. I switched from the explicit `Error!...` to an implicit error set to fix this.

I also see that the possible errors are listed in the documentation, but maybe it would be better to remove those and use an explicit error set (one that doesn't include `error.PoolIsFull`). Then `Error` could be defined as, for example, `error{PoolIsFull} || HandleError`. Let me know if you'd prefer that solution and I can update this PR :)

### DIFF
--- a/libs/zpool/src/pool.zig
+++ b/libs/zpool/src/pool.zig
@@ -1,7 +1,8 @@
 const std = @import("std");
 
-pub const PoolError = error{
-    PoolIsFull,
+pub const PoolError = error{PoolIsFull} || HandleError;
+
+pub const HandleError = error{
     HandleIsUnacquired,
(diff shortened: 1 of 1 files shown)


michal-z:
Yes, I would prefer to use explicit error sets. Thanks!

chip2n:
There we go, added explicit error set HandleError for a bunch of functions. I removed the part of the comments mentioning which errors are returned since the type signature should make that clear enough I hope :slightly_smiling_face:

michal-z:
Looks good, thanks!

PR #292, opened 2023-05-09
dxgi: Add support for ReportLiveObjects

gmodarelli:
Hello, I'm trying to add support for `ReportLiveObjects()`, but I'm getting a segmentation fault upon invocation and I can't figure out why.

I hope opening a PR might move things along :)

image

### DIFF
--- a/libs/zd3d12/src/zd3d12.zig
+++ b/libs/zd3d12/src/zd3d12.zig
@@ -609,6 +609,20 @@ pub const GraphicsContext = struct {
             _ = cmdalloc.Release();
         _ = gctx.wic_factory.Release();
         gctx.* = undefined;
+
+        if (enable_debug_layer) {
+            const debug = blk: {
+                var debug: *dxgi.IDebug1 = undefined;
+                hrPanicOnFail(dxgi.GetDebugInterface1(
+                    0,
+                    &dxgi.IID_IDebug1,
+                    @ptrCast(*?*anyopaque, &debug),
+                ));
+                break :blk debug;
+            };
+            _ = debug.ReportLiveObjects(dxgi.DEBUG_ALL, .{ .SUMMARY = true, .DETAIL = true, .IGNORE_INTERNAL = true });
+            _ = debug.Release();
+        }
     }
 
     pub fn beginFrame(gctx: *GraphicsContext) void {
--- a/libs/zd3d12/src/zd3d12.zig
+++ b/libs/zd3d12/src/zd3d12.zig
@@ -51,7 +51,7 @@ pub const GraphicsContext = struct {
     const max_num_buffered_resource_barriers = 16;
 
     device: *d3d12.IDevice9,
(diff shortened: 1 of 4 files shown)


michal-z:
I'm not sure but I think that this is old D3D11 API, for D3D12 we should be using https://learn.microsoft.com/en-us/windows/win32/api/d3d12sdklayers/nf-d3d12sdklayers-id3d12debugdevice-reportlivedeviceobjects

gmodarelli:
Perfect, thanks for the info. I'll update the PR :)

gmodarelli:
I think I've managed to get it to work. I've simulated a leak by not deleting a fence and now I get this message in the debug output window 

image

libs/zd3d12/src/zd3d12.zig
```diff
@@ -608,6 +621,11 @@ pub const GraphicsContext = struct {
         for (gctx.cmdallocs) |cmdalloc|
             _ = cmdalloc.Release();
         _ = gctx.wic_factory.Release();
+
+        if (enable_debug_layer) {
```
michal-z, line 625:
```suggestion
        if (gctx.debug_device) |debug_device| {
```

libs/zd3d12/src/zd3d12.zig
```diff
@@ -51,6 +51,7 @@ pub const GraphicsContext = struct {
     const max_num_buffered_resource_barriers = 16;
 
     device: *d3d12.IDevice9,
+    debug_device: *d3d12.IDebugDevice,
```
michal-z, line 54: Let's make it an optional (`?*d3d12.IDebugDevice`) and set it to `null` when debug layer is disabled. In that way users of `zd3d12` module will know if they can use this field.

michal-z (changes requested):
Thnaks! I've added few comments.

gmodarelli:
Thanks for the review!

libs/zd3d12/src/zd3d12.zig
```diff
@@ -189,6 +190,16 @@ pub const GraphicsContext = struct {
             break :blk device;
         };
 
+        var debug_device: *d3d12.IDebugDevice = undefined;
```
michal-z, line 193:
```suggestion
        var debug_device: ?*d3d12.IDebugDevice = null;
```

libs/zd3d12/src/zd3d12.zig
```diff
@@ -539,6 +550,7 @@ pub const GraphicsContext = struct {
 
         return GraphicsContext{
             .device = device,
+            .debug_device = if (enable_debug_layer) debug_device else null,
```
michal-z, line 553:
```suggestion
            .debug_device = debug_device,
```

michal-z (changes requested):
Just a super minor change. I think this is simpler because we have only one conditional now.

PR #294, opened 2023-05-13
zmath: allow for swapping near and far planes for reverse-z

### LINKED ISSUE #293: zmath: assert prevents reverse depth buffer implementation
Hello, 

I was implementing reverse depth buffer for our renderer and bumped into this assert from `zmath`
https://github.com/michal-z/zig-gamedev/blob/main/libs/zmath/src/zmath.zig#L2266

I need to pass inverted near and far to implement reverse depth buffer. For now I've patched `zmath` locally, but before contributing a PR I wanted to make sure that I was right :)

gmodarelli:
This PR fixes #293


michal-z:
Thanks!

PR #295, opened 2023-05-16
[zglfw] Expose waitEvents/waitEventsTimeout and fix test on X11

gviegas:
Changes two things:
- Exposes `glfwWaitEvents` and `glfwWaitEventsTimeout`
- Makes test more likely to succeed for non-blocking window setters (calls to `glfwSetWindowSize` and `glfwSetWindowPos` are asynchronous on X11)

The test was failing on both native Xorg and Xwayland.


michal-z:
Thanks for the PR!

PR #297, opened 2023-05-17
zphysics: Added a few missing API elements.

Srekel:
I'll likely add more later on, but wanted to get this pushed. 

Added functions:
**JPC_BodyInterface_GetPosition**
**JPC_BodyInterface_GetRotation**

Also added the **JPC_PhysicsUpdateError** return value to **JPC_PhysicsSystem_Update**.

### DIFF
--- a/libs/zphysics/libs/JoltC/JoltPhysicsC.h
+++ b/libs/zphysics/libs/JoltC/JoltPhysicsC.h
@@ -62,7 +62,7 @@ enum
 typedef uint8_t JPC_PhysicsUpdateError;
 enum
 {
-    JPC_PHYSICS_UPDATE_ERROR_NONE               = 0,
+    JPC_PHYSICS_UPDATE_SUCCESS               = 0,
     JPC_PHYSICS_UPDATE_MANIFOLD_CACHE_FULL      = 1 << 0,
     JPC_PHYSICS_UPDATE_BODY_PAIR_CACHE_FULL     = 1 << 1,
     JPC_PHYSICS_UPDATE_CONTACT_CONSTRAINTS_FULL = 1 << 2,
--- a/libs/zphysics/libs/JoltC/JoltPhysicsC.h
+++ b/libs/zphysics/libs/JoltC/JoltPhysicsC.h
@@ -1166,9 +1166,17 @@ JPC_BodyInterface_GetPointVelocity(const JPC_BodyInterface *in_iface,
                                    const JPC_Real in_point[3],
                                    float out_velocity[3]);
 JPC_API void
+JPC_BodyInterface_GetPosition(const JPC_BodyInterface *in_iface,
+                              JPC_BodyID in_body_id,
+                              JPC_Real out_position[3]);
+JPC_API void
 JPC_BodyInterface_GetCenterOfMassPosition(const JPC_BodyInterface *in_iface,
                                           JPC_BodyID in_body_id,
                                           JPC_Real out_position[3]);
+JPC_API void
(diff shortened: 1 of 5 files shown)


libs/zphysics/libs/JoltC/JoltPhysicsC.h
```diff
@@ -59,6 +59,15 @@ enum
     JPC_MAX_PHYSICS_BARRIERS = 8
 };
 
+typedef uint8_t JPC_PhysicsUpdateError;
```
michal-z, line 62:
```suggestion
typedef uint8_t JPC_PhysicsUpdateResult;
```

libs/zphysics/libs/JoltC/JoltPhysicsC.h
```diff
@@ -59,6 +59,15 @@ enum
...
+enum
+{
+    JPC_PHYSICS_UPDATE_ERROR_NONE               = 0,
```
michal-z, line 65:
```suggestion
    JPC_PHYSICS_UPDATE_SUCCESS               = 0,
```

libs/zphysics/libs/JoltC/JoltPhysicsC.h
```diff
@@ -1166,10 +1175,18 @@ JPC_BodyInterface_GetPointVelocity(const JPC_BodyInterface *in_iface,
...
+                              JPC_BodyID in_body_id,
+                              JPC_Real out_position[3]);
+JPC_API void
 JPC_BodyInterface_GetCenterOfMassPosition(const JPC_BodyInterface *in_iface,
                                           JPC_BodyID in_body_id,
                                           JPC_Real out_position[3]);
 JPC_API void
+JPC_BodyInterface_GetRotation(const JPC_BodyInterface *in_iface,
+                              JPC_BodyID in_body_id,
+                              JPC_Real out_rotation[4]);
```
michal-z, line 1188: Rotation always uses floats.
```suggestion
                              float out_rotation[4]);
```

michal-z (changes requested):
Thanks!

Few minor things.
* Rotations always use single precision so we need to use `float` instead of `JPC_Real`
* Let's rename JPC_PhysicsUpdateError -> JPC_PhysicsUpdateResult

libs/zphysics/libs/JoltC/JoltPhysicsC.h
Srekel: This is fine with me (that's what I would have named it too :) ), but just wanted to confirm given that the original enum is called Error:

external\zig-gamedev\libs\zphysics\libs\Jolt\Physics\EPhysicsUpdateError.h

/// Enum used by PhysicsSystem to report error conditions during the PhysicsSystem::Update call. This is a bit field, multiple errors can trigger in the same update.
enum class EPhysicsUpdateError : uint32
{
	None					= 0,			///< No errors
	ManifoldCacheFull		= 1 << 0,		///< The manifold cache is full, this means that the total number of contacts between bodies is too high. Some contacts were ignored. Increase inMaxContactConstraints in PhysicsSystem::Init.
	BodyPairCacheFull		= 1 << 1,		///< The body pair cache is full, this means that too many bodies contacted. Some contacts were ignored. Increase inMaxBodyPairs in PhysicsSystem::Init.
	ContactConstraintsFull	= 1 << 2,		///< The contact constraints buffer is full. Some contacts were ignored. Increase inMaxContactConstraints in PhysicsSystem::Init.
};

michal-z:
@Srekel I did minor tweaks to the PR.

PR #299, opened 2023-05-18
[zjobs] Make throughput test work on systems with 2-4 GiB of RAM

gviegas:
This fix makes  `test "JobQueue throughput"`  use less jobs in such systems.


michal-z:
Thanks for the PR!

PR #301, opened 2023-05-25
zglfw: add setClipboardString, getClipboardString, and setDropCallback

xdBronch:
i was unsure where to put the clipboard functions as they take a ``window`` parameter but glfw docs say that its deprecated and can also just take null. not familiar with the standard way of handling strings with extern C functions so ill gladly change something if needed.

### DIFF
--- a/libs/zglfw/src/zglfw.zig
+++ b/libs/zglfw/src/zglfw.zig
@@ -596,6 +596,15 @@ pub const Window = opaque {
     }
     extern fn glfwSetWindowTitle(window: *Window, title: [*:0]const u8) void;
 
+    pub const getClipboardString = glfwGetClipboardString;
+    extern fn glfwGetClipboardString(window: *Window) ?[*:0]const u8;
+
+    pub const setClipboardString = glfwSetClipboardString;
+    extern fn glfwSetClipboardString(
+        window: *Window,
+        string: [*:0]const u8,
+    ) void;
+
     /// `pub fn setFramebufferSizeCallback(window: *Window, callback) void`
     pub const setFramebufferSizeCallback = glfwSetFramebufferSizeCallback;
     extern fn glfwSetFramebufferSizeCallback(window: *Window, callback: ?*const fn (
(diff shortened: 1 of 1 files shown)


xdBronch:
> let's use `[:0]const u8` for user facing Zig functions.

sure i can definitely change it but that only works when taking in a string and not returning it like with ``getClipboardString``, is that fine? just leave as is?

michal-z:
When returning a strint let's use `std.mem.span()` like below:

```zig
    pub fn getGuid(self: Joystick) [:0]const u8 {
        return std.mem.span(glfwGetJoystickGUID(@intCast(i32, self.jid)));
    }
    extern fn glfwGetJoystickGUID(jid: i32) [*:0]const u8;
```

Thanks!

libs/zglfw/src/zglfw.zig
```diff
@@ -596,6 +596,19 @@ pub const Window = opaque {
     }
     extern fn glfwSetWindowTitle(window: *Window, title: [*:0]const u8) void;
 
+    pub fn getClipboardString(window: *Window) ?[:0]const u8 {
+        return std.mem.span(glfwGetClipboardString(window));
```
michal-z, line 600:
```suggestion
        return if (glfwGetClipboardString(window)) |str| std.mem.span(str) else null;
```

xdBronch:
sounds good, done

michal-z (changes requested):
I think that `getClipboardString()` won't compile I've added proposed fix (not tested).

libs/zglfw/src/zglfw.zig
xdBronch: not gonna pretend i knew this before lol but it actually does work.
```zig
pub fn span(ptr: anytype) Span(@TypeOf(ptr)) {
    if (@typeInfo(@TypeOf(ptr)) == .Optional) {
        if (ptr) |non_null| {
            return span(non_null);
        } else {
            return null;
        }
...
```
from the stdlib

michal-z: Oh, nice!

Issue #284, opened 2023-04-18
zflecs: Components cannot contain `zmath.F32x4`


foxnne:
Not sure if this should be obvious, but I believe zig helped catch something that was not obvious to me before: vectors are not supported as a component field. The error I was running into was pointer misalignment during `ecs.field()` calls. Does this mean that only `extern` aligned structs are allowed as components? Or are vectors a special case?

michal-z:
Will take a look.

Note that it is better to use something like `[2]f32` in a component and load the data only we you need it with `loadArr2()`. `zmath.F32x4` type is an abstraction for a CPU register and it should be used as a temporary storage for performing computations.

foxnne:
Understood! Thanks for the explanation!

michal-z:
That said, it should still work. Re-opening for the investigations :)

PR #303, opened 2023-05-27
zflecs: fix wrong alignment issue #284

LukasKastern:
[This](https://github.com/SanderMertens/flecs/issues/478.) is the flecs issue for reference.
Compiling flecs using ```-DFLECS_USE_OS_ALLOC``` already works around the issue. 

But I've also implemented the allocation related os calls using a gpa to track memory leaks and guarantee correct alignment. 
I haven't benchmarked the performance impact of using it. So it might make sense to make it optional 🤷.

### DIFF
--- a/libs/zflecs/build.zig
+++ b/libs/zflecs/build.zig
@@ -30,6 +30,7 @@ pub fn package(
     zflecs_c_cpp.addCSourceFile(thisDir() ++ "/libs/flecs/flecs.c", &.{
         "-fno-sanitize=undefined",
         "-DFLECS_NO_CPP",
+        "-DFLECS_USE_OS_ALLOC",
         if (@import("builtin").mode == .Debug) "-DFLECS_SANITIZE" else "",
     });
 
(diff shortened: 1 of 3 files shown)


LukasKastern:
> So, this adds 128 bytes to each allocation, right? This seems expensive to me.

Oh sorry. I mixed bits/bytes up in this pull request. The max alignment should be 16 bytes not 128 lol.
I'll fix it up later today 😅.

Do you think 16 bytes of extra space for the allocation header is fine or should we use a hashmap anyways?

michal-z:
Maybe we could use 8 bytes for a header and store there size (`u48`) and alignment (`u16`) that would be the best solution.

LukasKastern:
> Maybe we could use 8 bytes for a header and store there size (u48) and alignment (u16) that would be the best solution.

That would also mean that the maximum alignment that flecs can request is gonna be 8 bytes. 
I'm gonna try that out and see if it works.

[Glibc malloc](https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/malloc/malloc.c#L98) seems to align to 16 bytes on 64bit platforms by default. 

I guess we can expose the alignment as a build_option in case you have a component that requires a greater alignment.

LukasKastern:
Should be good now. Let me know if you would like me to change something.

PR #308, opened 2023-05-31
zgui: Remove `usingnamespace` to fix zls

foxnne:
I created an [issue](https://github.com/zigtools/zls/issues/1218) on the zls page, but I'm not sure how long it will take for `usingnamespace` issues to be resolved. Personally, I find zls really helpful especially with zgui. The fix here could just be temporary until things are fixed on the zls side. If you don't want to make this change I can just make the change on my own branch. Thanks!


libs/zgui/src/gui.zig
```diff
@@ -1,4 +1,19 @@
 //--------------------------------------------------------------------------------------------------
+//
+// Zig bindings for 'dear imgui' library. Easy to use, hand-crafted API with default arguments,
+// named parameters and Zig style text formatting.
+//
+//--------------------------------------------------------------------------------------------------
+pub const version = @import("std").SemanticVersion{ .major = 0, .minor = 9, .patch = 6 };
+
+pub usingnamespace @import("gui.zig");
```
michal-z, line 9: Does this work? File exports itself?

foxnne: Whoops! I meant to remove that line, just pushed the change.

PR #224, opened 2023-02-24
emscripten PoC

hazeycode:
Hacky PoC to illustrate horrible workarounds required to support emscripten and start a discussion on the topic.

Currently, this builds and then hits some unreachable at runtime. I will do some debugging and try to get it further if and when I can but no promises.

To build and run the minimal_sdl sample in your web browser:
`zig build minimal_sdl-run -Dtarget=wasm32-emscripten --sysroot emscripten/emsdk/upstream/emscripten/cache/sysroot`

required reading:
- https://github.com/ziglang/zig/issues/10836

todo: 
- request webgl compatible context

PR #309, opened 2023-05-31
Emscripten support: wasm & webgpu in browser

Deins:
Feel free to reject this if you think this is out of scope or if some of the required workarounds are too ugly etc. But I thought to share this in case this interesting/useful and hear what you think.

This adds support and a sample for web compilation with `emscripten` toolchain. In general it just adds additional linking step with emsdk on top of regular `wasm32-freestanding` build which generates all the nececary js browser binding glue code. But also needs some platform specific modifications.
Main notes with required changes are in `triangle_wgpu_emscripten` sample readme.
Live demo (needs nightly browser with webgpu support): https://deins.lv/triangle_wgpu_emscripten/.

Its quite raw and its possible that some of the emscripten specific stuff could be better abstracted and included in zgpu for example. But currently I tried to avoid modifying api in any way requiring user/sample_code to do it instead. 

Known issues
* uses stock html shell that doesn't display any user friendly message if browser doesn't support webgpu
* not all keyboard input works correctly in imgui, such as arrow keys / backspace. Might be glfw version incompatibility, has to be debugged.
* mapAsync is much slower and there is larger chance of running out of uniform buffers. Blocking is not viable and requires added `canRender` fn to be called before each frame to prevent crashes in such scenarios. Or in most extreme situations skipping frame. Also requires emscripten from upstream as I had to make pull request in regards to mapAsync alignment.


michal-z:
Thanks for the PR and nice work!

I will be looking at this in coming days but please note that I have zero experience with the web stuff. That said, I would like to have this functionality in the project.

@hazeycode Can you please help with the review?

Also, how does this compare to https://github.com/michal-z/zig-gamedev/pull/224?

hazeycode:
Very nice work @Deins !

@michal-z This is focused on linking glfw-webgpu using emscipten where in #224 I tried linking sdl-webgl. Also the way the program entry point is resolved is different. I need to revisit some of the details. I will try to review before the end of the week.

Interesting that Firefox is not playing along. That's what I was testing with last time I looked at this!

Deins:
Sorry, missed #224. It is similar, but yes main difference is glfw vs sdl and webgl vs webgpu. 

I don't see a reason why they both  could not work. As in this project user can choose which libs to use and which backend. Main requirement would be to create a nice way from build system to be able to choose correct emscripten link flags, etc.

Another difference is that this uses `wasm32-freestanding` target instead of `wasm32-emscripten`. Reason is: I ran into issues with zig `std` that had almost no support for emscripten os tag even for most basic things. And it might be that it is the reason why the c entry point was needed for #224. Overall there is no need to use `wasm32-emscripten` its only needed for C/C++ libraries that use emscripten macro magic to inline JS in C source files. Otherwise `freestanding` for zig code is much nicer and I don't know any downsides other than not being able to use builtin `os.tag` for platform specific checks.  So most ergonomic build flow from my experience is to:
* use `freestanding` for zig code
* use `freestanding` for C libraries that have no emscripten specific macros
* use `emscripten` target tag for C libraries that use emscripten specific macros

In regards to firefox, I got webgpu running on different pc. But it still didn't run: webgpu function [onSubmittedWorkDone](https://developer.mozilla.org/en-US/docs/Web/API/GPUQueue/onSubmittedWorkDone), is not implemented in it. I am not sure how critical it is in zgpu, probably can be worked around. But it  might get implemented in future so might not need to worry too much about it until any browser actually enables webgpu by default in stable releases and doesn't have it (as from my understanding it is in webgpu spec).

hazeycode:
> I don't see a reason why they both could not work. As in this project user can choose which libs to use and which backend. Main requirement would be to create a nice way from build system to be able to choose correct emscripten link flags, etc.

Agreed

> Overall there is no need to use wasm32-emscripten its only needed for C/C++ libraries that use emscripten macro magic to inline JS in C source files. Otherwise freestanding for zig code is much nicer and I don't know any downsides other than not being able to use builtin os.tag for platform specific checks. So most ergonomic build flow from my experience is to:
>     use freestanding for zig code
>     use freestanding for C libraries that have no emscripten specific macros
>     use emscripten target tag for C libraries that use emscripten specific macros

Also agreed. IIRC SDL makes use of these macros and there is some complication with how the main shim works. New SDL 3 simplifies this FWIW. Regardless, we can focus on GLFW+WebGPU here.

build.zig
```diff
@@ -482,3 +501,75 @@ fn ensureGitLfsContent(comptime file_path: []const u8) !void {
...
+    }
+    const out_path = b.pathJoin(&.{ b.pathFromRoot("."), "zig-out", "www", exe.name });
+    std.fs.cwd().makePath(out_path) catch unreachable;
+    const out_file = try std.mem.concat(b.allocator, u8, &.{ "-o", out_path, std.fs.path.sep_str ++ "index.html" });
+    emlink.addArgs(&.{"-sEXPORTED_FUNCTIONS=['_malloc','_emmalloc_realloc','_main']"});
+    //emling.addArgs(&.{"-sLLD_REPORT_UNDEFINED", "-sERROR_ON_UNDEFINED_SYMBOLS=0"});
+    emlink.addArg(out_file);
+    emlink.addArg("-sFILESYSTEM=0");
+    emlink.addArg("-sUSE_GLFW=3");
+    emlink.addArg("-sUSE_WEBGPU=1");
```
hazeycode, line 532: Library/sample specific flags could be added separately i.e. `linkEmscripten` only has the common stuff

libs/zglfw/build.zig
```diff
@@ -3,12 +3,15 @@ const std = @import("std");
...
     pub fn link(pkg: Package, exe: *std.Build.CompileStep) void {
         exe.addModule("zglfw", pkg.zglfw);
 
         const host = (std.zig.system.NativeTargetInfo.detect(exe.target) catch unreachable).target;
 
+        if (pkg.options.emscripten) return;
+
         switch (host.os.tag) {
             .windows => {},
             .macos => {
```
hazeycode, line 17: We can just leverage the Zigs builtin emscripten target here and treat it like the others.

libs/zgpu/src/zgpu.zig
```diff
@@ -345,6 +368,11 @@ pub const GraphicsContext = struct {
         }
 
         if (gctx.uniforms.stage.num >= uniforms_staging_pipeline_len) {
+            if (emscripten) {
+                // we can't block in requestAnimationFrame
+                slog.debug("uniformsNextStagingBuffer: Out of buffers! canRender() must be checked next frame, otherwise we will crash!", .{});
```
hazeycode, line 373: This should be `slog.warning`

samples/triangle_wgpu_emscripten/src/triangle_wgpu.zig
```diff
@@ -0,0 +1,550 @@
...
+        var buffer: [1024]u8 = undefined;
+        const path = std.fs.selfExeDirPath(buffer[0..]) catch ".";
+        std.os.chdir(path) catch {};
+    }
+
+    if (emscripten) {
+        // by default emscripten initializes on window creation WebGL context
+        // this flag skips context creation. otherwise we later can't create webgpu surface
+        zglfw.WindowHint.set(.client_api, @enumToInt(zglfw.ClientApi.no_api));
+    }
```
hazeycode, line 361: Should this just be set anyway?

samples/triangle_wgpu_emscripten/src/triangle_wgpu.zig
```diff
@@ -0,0 +1,550 @@
...
+
+    pub export fn glfwGetGamepadState(_: i32, _: ?*anyopaque) i32 {
+        return 0; // false - failure
+    }
+
+    pub export fn wgpuDeviceTick() void {
+        std.log.warn("use of device.tick() should be avoided! It can break if used with callbacks such as requestAnimationFrame etc.", .{});
+        emscripten_sleep(1); // requires -sASYNCIFY
+    }
+} else struct {};
```
hazeycode, line 550: Can we move this into `zglfw`?

samples/triangle_wgpu_emscripten/src/triangle_wgpu.zig
```diff
@@ -0,0 +1,550 @@
...
+    if (!emscripten) {
+        var buffer: [1024]u8 = undefined;
+        const path = std.fs.selfExeDirPath(buffer[0..]) catch ".";
+        std.os.chdir(path) catch {};
+    }
+
+    if (emscripten) {
+        // by default emscripten initializes on window creation WebGL context
+        // this flag skips context creation. otherwise we later can't create webgpu surface
+        zglfw.WindowHint.set(.client_api, @enumToInt(zglfw.ClientApi.no_api));
```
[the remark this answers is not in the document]
Deins: not sure, if someone wants to use webgl +glfw, different flags might be needed, but I haven't tested

libs/zgpu/src/binding_tests.zig
```diff
@@ -0,0 +1,84 @@
...
+const gpu = @import("wgpu.zig");
+
+// test struct sizes
+fn assertStructBindings(zig_type: anytype, c_type: anytype) void {
+    if (@sizeOf(zig_type) != @sizeOf(c_type)) {
+        @compileLog("emscripten zgpu type has different size from webgpu.h type:", zig_type, c_type, @sizeOf(zig_type), @sizeOf(c_type));
+        unreachable;
+    }
+    // more checks? iterate struct fields?
+    // emscripten in .Debug mode asserts at runtime if required field is missing etc. but something always slips through
```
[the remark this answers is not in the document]
Deins: core issue is why I added these tests is that `webgpu.zig` basically rewrites c header `webgpu.h`. By default it matches dawn `webgpu.h` but emscripten ships with its own that can have version mismatch. If zig structs doesn't match with ones in actual header problems start. For emscripten that means it will try read some struct field but might get value from different field etc. Size checks help to quickly at comptime show these. But if u32 field was removed to struct and other added at end, it will slip through leading to unknown behavior and most likely crash at runtime or worse. But its not straight forward to test at comptime whole struct due to camel-case/snake-case naming changes.

libs/zglfw/build.zig
Deins: Just commited this. Not sure if its better. Its still tricky because zig modules reuse target from target its linked with (as I understand), now zglfw and zgpu etc. has to assume both .emscripten and .freestanding target means emscripten. It works as long as we don't need any other .freestanding target. But still we have to  support `wasm32-freestanding` and `emscripten` os, or beter workaround has to be found to be able to use `.emscripten` target everywhere.

Deins:
Fixed some of the requested changes, still might need some cleanup. 
* removed/ switched to target `wasm32-emscripten`. 
* separated / moved  emscripten build and utility stuff in separate lib `zems`. I tried to avoid adding it as dependency to zglfw, zgpu etc. So that mostly stays the same. 
* ported existing `gui_test_wgpu` to see how easy it is to port existing sample with refactored build stuff [demo](https://deins.lv/gui_test_wgpu/). Includes usage of `emscripten` asset file preloading that worked well. But file ops are mostly called from c, haven't tested how / if std.fs would work from zig code. Most likely would require some hacks to make std.fs to use posix filesystem calls in either .emscripten or .freestanding target (has to be investigated).

build.zig
```diff
@@ -95,28 +97,32 @@ fn packagesCrossPlatform(b: *std.Build, options: Options) void {
...
     zgpu_pkg = zgpu.package(b, target, optimize, .{
         .options = .{ .uniforms_buffer_size = 4 * 1024 * 1024 },
         .deps = .{ .zpool = zpool_pkg.zpool, .zglfw = zglfw_pkg.zglfw },
     });
+    zems_pkg = zems.package(b, target, optimize, .{});
     ztracy_pkg = ztracy.package(b, target, optimize, .{
         .options = .{
-            .enable_ztracy = !target.isDarwin(), // TODO: ztracy fails to compile on macOS.
-            .enable_fibers = !target.isDarwin(),
+            .enable_ztracy = !target.isDarwin() and options.target.getOsTag() != .emscripten, // TODO: ztracy fails to compile on macOS.
```
hazeycode, line 113: @michal-z is this comment still valid? I think we fixed it here: https://github.com/michal-z/zig-gamedev/pull/262?

libs/zgpu/src/binding_tests.zig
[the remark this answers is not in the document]
hazeycode: @michal-z do you have any thoughts regarding how struct layouts can be validated?

samples/triangle_wgpu_emscripten/src/triangle_wgpu.zig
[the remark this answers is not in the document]
hazeycode: this is a wgpu sample so I think it's ok

libs/zems/zems.zig
```diff
@@ -0,0 +1,119 @@
...
+    return c.emscripten_request_animation_frame_loop(cb_ptr, user_data);
+}
+
+pub inline fn cancelAnimationFrame(request_animation_frame_id: c_long) void {
+    c.emscripten_cancel_animation_frame(request_animation_frame_id);
+}
+
+pub inline fn sleep(ms: u32) void {
+    c.emscripten_sleep(ms);
+}
```
hazeycode, line 36: I think we should just declare extern C fns instead of wrapping them. What do you think @michal-z ?

libs/zglfw/build.zig
hazeycode: > now zglfw and zgpu etc. has to assume both .emscripten and .freestanding target means emscripten

Why? I don't understand. Do you mean that if we compile our game as freestanding we are unable to link zglfw that has been built with emscripten as os? I thought that it would link because they are both wasm32.

samples/gui_test_wgpu/src/gui_test_wgpu.zig
```diff
@@ -579,40 +585,65 @@ fn draw(demo: *DemoState) void {
...
+        if (demo.gctx.canRender()) tick(demo) catch |err| {
+            std.log.err("animation frame canceled! tick failed with: {}", .{err});
+            return .false;
+        } else {
+            std.log.warn("canRender(): Frame skipped!", .{});
+        }
+        return .true;
+    }
+} else struct {};
+extern fn tickEmcripten(time: f64, user_data: ?*anyopaque) callconv(.C) zems.EmBool;
```
hazeycode, line 649: I think we should look into abstracting this away in zglfw. But I'm not sure exactly what it would look like.... something like zglfw init sets up a js animation callback which sets some redraw bool that we can check in our application loop. It's not the ideal way of handling browser gpu sync but I think it's "good enough". Tight gpu synchronization requires using a better platform layer than what glfw provides anyway.

It would of course mean than zems becomes an optional dependency of zglfw, which I think is fine.

What do you guys think? @Deins @michal-z

libs/zems/dummy.zig
```diff
@@ -0,0 +1,7 @@
+// dummy interface for non emscripten builds
+pub const is_emscripten = false;
+
+pub const EmBool = enum(c_int) {
+    true = 1,
+    false = 0,
+};
```
hazeycode, line 7: non-emscripten builds shouldn't have zems at all

libs/zems/zems.zig
```diff
@@ -0,0 +1,119 @@
+pub const is_emscripten = true;
```
hazeycode, line 1: If user is linking zems; either directly or by enabling emsc support in zglfw then we should be able to detect that without this

hazeycode (changes requested):

hazeycode:
> I tried to avoid adding it as dependency to zglfw, zgpu etc. So that mostly stays the same.

As I alluded to in review comments, I don't see why zglfw and zgpu should avoid emsc glue dependency, in fact I think it's a good idea for them to *optionally* depend on emsc glue so that they may abstract away platform details, which is their purpose.

samples/gui_test_wgpu/src/gui_test_wgpu.zig
Deins: I think its not glfws fault. (at least for web) 
Sdl2 had the same issue a while back, it required user to use emscripten specific callbacks for web (at least in older versions that I remember). Random example from web: https://github.com/erik-larsen/emscripten-sdl2-ogles2/blob/master/src/hello_triangle.cpp#L144

Core issue is that on web nothing can block main thread. (and `WebWorker` threads can't access any web api, so its not even worth mentioning). Emscripten provides `asyncify` magic that will try to detect functions that can block and do some crazy stuff to unwind stack etc. to allow blocking code to actually not block. But it has its limitations.
In general to smoothly render there is no alternative as far as I know than using requestAnimationFrame.

I think unless API is ground up built, not to ever block, it doesn't matter, it will require getting rid of main loop and using web specific callback interface. Nice example of how such api might look like is C library [sokol](https://github.com/floooh/sokol) which just tries to provide cross platform window, input etc.
But I think its out of scope of `zig-gamedev` which mostly just provides zig bindings and libraries which you match together as you wish.

So yeah,  currently maybe is good idea just to have one example that show how to compile for web with emscripten.

hazeycode: Yeah you're right, the problem is trying to fudge a desktop platform abstraction to work with browsers and they we're never designed to do so.

In my own projects I use a custom platform abstraction with completely separate desktop and web backends. Maybe when it is more mature I will look at contributing the web portion to zig-gamedev. The use-case here I think is being to quickly be able to take something written on-top of glfw and get it going in a browser. Which wouldn't the path I would recommend for any serious project. But it's a usecase that *can* be supported nonetheless.

> So yeah, currently maybe is good idea just to have one example that show how to compile for web with emscripten.

Yeah I think this is the way to go, for now it can just be glfw+webgpu. But in the future we could potentially reuse `zems` to show how to get sdl+opengl in the browser too. Sorry for the distraction by suggesting that we should try to get existing samples to work.

Let's see if @michal-z agrees

michal-z: Yes, I agree. One sample compiled for the web is enough for this PR. It will provide nice example and give us experience/knowledge about what exactly is needed - we can think about better/different abstraction in the future.

libs/zglfw/build.zig
Deins: Basically issue is that I had to use `wasm32-freestanding` instead of emscripten for main user/sample code, otherwise I get std lib errors such as:
```
zig-master\lib\std\os.zig:153:24: error: root struct of file 'c' has no member named 'fd_t'
```

In regards to zglfw & libs issue is not with `c` but zig modules that don't support specifying different target for them: [build.createModule](https://ziglang.org/documentation/master/std/#A;std:Build.createModule) - can't specify target. [addModule()](https://ziglang.org/documentation/master/std/#A;std:Build.addModule) just uses target from exe its added to. 
So zig bindings code such as `zglfw.zig` in end has to assume that `builtin.target` freestanding is emscripten.
Hope I explained it better this time.

hazeycode: Yes, i follow now. Thanks for clarifying. I'm not sure whether it's better to deal with the entry point or not in the long run; but let's go with what works for you for now.

samples/triangle_wgpu_emscripten/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Build
+#### Native
+```bash
+zig build triangle_wgpu_emscripten-run
+```
+#### Wasm for web
+#### Install emscripten sdk
+* [Follow these instructions](https://emscripten.org/docs/getting_started/downloads.html#installation-instructions-using-the-emsdk-recommended)
+* emscripten needs this [patch](https://github.com/emscripten-core/emscripten/pull/19477/commits/f4bb4f578131578cd13abbbf78d7f4273788d76f) currently for this sample to run until it gets merged/released.
```
hazeycode, line 13: This has now been merged :champagne:

samples/triangle_wgpu_emscripten/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+#### Compile
+```bash
+zig build triangle_wgpu_emscripten -Dtarget=wasm32-emscripten
```
hazeycode, line 17: This is currently failing for me:
```
zig build-lib zgui Debug wasm32-emscripten: error: error(compilation): clang failed with stderr: /home/hazeycode/Desktop/zig-gamedev/libs/zgui/libs/imgui/backends/imgui_impl_wgpu.cpp:337:15: error: no member named 'source' in 'WGPUShaderModuleWGSLDescriptor'
/home/hazeycode/Desktop/zig-gamedev/libs/zgui/libs/imgui/backends/imgui_impl_wgpu.cpp:606:37: error: assigning to 'WGPUMipmapFilterMode' from incompatible type 'const WGPUFilterMode'


zig build-lib zgui Debug wasm32-emscripten: error: the following command failed with 1 compilation errors:
/home/hazeycode/zig/0.11.0-dev.4010+70c71935c/files/zig build-lib -cflags -fno-sanitize=undefined -- /home/hazeycode/Desktop/zig-gamedev/libs/zgui/src/zgui.cpp -cflags -fno-sanitize=undefined -- /home/hazeycode/Desktop/zig-gamedev/libs/zgui/libs/imgui/imgui.cpp -cflags -fno-sanitize=undefined -- /home/hazeycode/Desktop/zig-gamedev/libs/zgui/libs/imgui/imgui_widgets.cpp -cflags -fno-sanitize=undefined -- /home/hazeycode/Desktop/zig-gamedev/libs/zgui/libs/imgui/imgui_tables.cpp -cflags -fno-sanitize=undefined -- /home/hazeycode/Desktop/zig-gamedev/libs/zgui/libs/imgui/imgui_draw.cpp -cflags -fno-sanitize=undefined -- /home/hazeycode/Desktop/zig-gamedev/libs/zgui/libs/imgui/imgui_demo.cpp -cflags -fno-sanitize=undefined -- /home/hazeycode/Desktop/zig-gamedev/libs/zgui/libs/imgui/implot_demo.cpp -cflags -fno-sanitize=undefined -- /home/hazeycode/Desktop/zig-gamedev/libs/zgui/libs/imgui/implot.cpp -cflags -fno-sanitize=undefined -- /home/hazeycode/Desktop/zig-gamedev/libs/zgui/libs/imgui/implot_items.cpp -cflags -fno-sanitize=undefined -- /home/hazeycode/Desktop/zig-gamedev/libs/zgui/libs/imgui/backends/imgui_impl_glfw.cpp -cflags -fno-sanitize=undefined -- /home/hazeycode/Desktop/zig-gamedev/libs/zgui/libs/imgui/backends/imgui_impl_wgpu.cpp --cache-dir /home/hazeycode/Desktop/zig-gamedev/zig-cache --global-cache-dir /home/hazeycode/.cache/zig --name zgui -static -fno-stack-check -fno-stack-protector -target wasm32-emscripten -mcpu generic -I /home/hazeycode/Desktop/zig-gamedev/libs/zgui/libs -I /home/hazeycode/Desktop/zig-gamedev/libs/zgui/libs/imgui -isystem /home/hazeycode/Desktop/emsdk/upstream/emscripten/cache/sysroot/include -D __EMSCRIPTEN__ -D __EMSCRIPTEN_major__=3 -D __EMSCRIPTEN_minor__=1 --listen=- 
Build Summary: 3/7 steps succeeded; 1 failed (disable with --summary none)
triangle_wgpu_emscripten transitive failure
└─ run emcc transitive failure
   ├─ zig build-lib triangle_wgpu_emscripten Debug wasm32-freestanding transitive failure
   │  └─ zig build-lib zgui Debug wasm32-emscripten 1 errors
   ├─ zig build-lib zgui Debug wasm32-emscripten (reused)
   └─ zig build-lib triangle_wgpu_emscripten Debug wasm32-freestanding (+6 more reused dependencies)
/home/hazeycode/Desktop/zig-gamedev/libs/zgui/libs/imgui/backends/imgui_impl_wgpu.cpp:1:1: error: unable to build C object: clang exited with code 1
```

Note I am able to compile for native host target. This is only a problem when compiling for `wasm32-emscripten`

hazeycode:
Needs a rebase

ckrowland:
Hello everyone.

I'm not sure exactly how to do this, but I rebased deins:emscripten onto michal-z:main on my branch ckrowland:em. Hopefully it can save someone some time.
If my attempt is acceptable, Should I create a PR on deins:emscripten branch or create a new pull request here on michal-z:main?

I got all the emscripten examples running on the web with
chrome, zig 0.11.0-dev.4403+e84cda0eb, macos 12.6.7 and the current zig-gamedev main.

Firefox nightly is throwing a 
`queue.onSubmittedWorkDone is not a function` error

What else needs to be done? I'm still getting up to speed on emscripten/wasm.

Screen Shot 2023-08-04 at 5 54 08 PM

michal-z:
@Deins Could you please rebase this? We can take it from there if you don't have time to work on it anymore. I think we are very close to finally merge it. Thanks for your work!

Deins:
sorry, have been too busy to work  much on this. 
Overall I think its in ok state. Since wgpu.zig was updated to newer version, as well `wgpuCreateInstance` implementation in emscripten, some of the ugly hacks are gone now. 
Just have to be careful with emscripten versions as wgpu still slightly changes.
Tested now with zig 11 and emscripten sdk from latest master commit `ef2a8e929d5337755e9b1d1e1d4ad859dc694eee` seems to be working.

The last thing that still might be nice to get rid of is zgpu `canRender` check that I added to safeguard against running out of buffers. 
Otherwise just testing it in more complex use cases. I think there might be more of minor usize/u64 mistakes in some wgpu functions but those should be easy to identify and fix.

hazeycode:
Some useful developments/info here: https://github.com/ziglang/zig/issues/10836#issuecomment-1703611299

libs/zems/build.zig
```diff
@@ -0,0 +1,179 @@
...
+                emlink.addArg(std.mem.join(b.allocator, "", &.{ "-s", opt.key_ptr.*, "=", opt.value_ptr.* }) catch unreachable);
+            } else {
+                emlink.addArg(std.mem.join(b.allocator, "", &.{ "-s", opt.key_ptr.* }) catch unreachable);
+            }
+        }
+
+        emlink.addArgs(self.args.other_args.items);
+
+        if (self.args.shell_file) |sf| emlink.addArgs(&.{ "--shell-file", sf });
+
```
ckrowland, line 99: To load-files with emscripten, I added
```if (self.args.content_dir) |cd| emlink.addArgs(&.{ "--preload-file", cd });```
here

File: libs/zems/build.zig [hunk 16]
```diff
@@ -0,0 +1,179 @@
...
+        emlink.step.dependOn(&exe.step);
+        self.link_step = emlink;
+        self.lib_exe = exe;
+    }
+};
+
+pub const EmscriptenArgs = struct {
+    source_map_base: []const u8 = "./",
+    shell_file: ?[]const u8 = null,
+
```
ckrowland, line 109: Then add the arg:
```content_dir: ?[]const u8 = "content",```

I found you can use files like you would natively. For loading fonts, images, etc.

ckrowland:
I was able to get my [demo](https://connorrowland.me/random/) running live on chromium browsers with this. So much thanks Deins!

Only a few minor things I found:
- Needed to pass -sUSE_OFFSET_CONVERTER to emsdk
- Needed to copy shell.html file emscripten relies on into zems
  - Maybe we should include a basic template to start?

hazeycode:
> Maybe we should include a basic template to start?

Yes I think this is a good idea!

hazeycode:
Having another go at emscripten support here -> https://github.com/zig-gamedev/zig-gamedev/pull/480, based largely on this PR

hazeycode:
Note that we have [one demo](https://github.com/zig-gamedev/zig-gamedev/tree/main/samples/sdl2_demo) building for web. I'm leaving this open to remind me to get the other demos, including wgpu ones in a browser too.

taylorh140:
> Note that we have [one demo](https://github.com/zig-gamedev/zig-gamedev/tree/main/samples/sdl2_demo) building for web. I'm leaving this open to remind me to get the other demos, including wgpu ones in a browser too.

I was curious to try this. (probably the wrong place to ask the question) but what are the build options activate this?

ckrowland:
Hey @taylorh140, emscripten is still a work in progress. I was able to get my own project working with all the pull requests I referenced. Once we get those merged you'll have better luck.

PR #313, opened 2023-06-13
zsdl: use canonical SDL names

hazeycode:
Small tweaks to some names so that they match SDL. This makes it easier to work with when referencing SDL documentation.


Jack-Ji:
I’m ok with it. Thanks for heads up BTW.

Issue #314, opened 2023-06-13
zsdl: Cannot declare undefined PixelFormat


hazeycode:
Trying to declare an undefined PixelFormat produces the following error:
```
error: reached unreachable code
    if (!ok) unreachable; // assertion failure
             ^~~~~~~~~~~
libs/zsdl/src/zsdl.zig:328:19: note: called from here
            assert(@TypeOf(order) == BitmapOrder);
            ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
libs/zsdl/src/zsdl.zig:345:31: note: called from here
    index8 = definePixelFormat(.index8, 0, 0, 8, 1),
             ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
```

To reproduce drop the following in any code that uses zsdl:
```zig
var pixel_format: sdl.PixelFormat = undefined;
_ = pixel_format;
```

Tested with Zig 0.11.0-dev.3395+1e7dcaa3a

Any ideas @Jack-Ji ?

Jack-Ji:
PR #315 has been created to fix the issue.

PR #315, opened 2023-06-14
fix issue #314

Jack-Ji:
Fix definition of PixelFormat


hazeycode (approved):
ACK: I have tested this with Zig 0.11.0-dev.3395+1e7dcaa3a and confirm that it fixes #314, thanks @Jack-Ji !

Issue #318, opened 2023-06-17
std.math.max, std.math.min deprecated


kamidev:
Breaking zig change: https://github.com/ziglang/zig/pull/16025

std.math.max -> @max,  std.math.min -> @min

A simple search and replace is sufficient to make zig-gamedev build and pass all tests again.

michal-z:
Fixed in #319

PR #322, opened 2023-06-18
Fully fix issue 318: std.math.max, std.math.min deprecated

BoilerplateJoe:
I tried to compile zig-gamedev with the latest Zig version (`0.11.0-dev.3677+e95fc2023`) and got a lot of deprecation messages about `std.math.min` and `std.math.max`, which are now `@min` and `@max`. I saw this was a reported issue (#318) that got resolved by PR #319, but it looks like some were missed. This PR should hopefully resolve all of the deprecation messages.


michal-z:
Thanks!

PR #326, opened 2023-06-22
zpool: Fix int overflows

orls:
I found that `audio_experiments_wgpu` sample would crash after a few mins with an integer overflow:

- The iterators inside pool could overflow their loop counters when pool is full.
- The sample app itself was filling a pool due to not properly releasing handles.


michal-z:
Thanks!

PR #330, opened 2023-07-04
Add return values to GLFW set callback functions

dstrachan:
The GLFW backend for ImGui sets input callbacks which get discarded when subsequent calls are made to `window.set*Callback` functions. Adding the return values to these functions allows the caller to handle cases where the callbacks have already been set, without the need to reimplement the ImGui backend input handling.


michal-z:
Thanks!

Issue #328, opened 2023-06-30
zgui: dragpoints support


ssteinbach:
Hi, love this project!

Are there any plans to add `DragPoints` to the Implot wrapper?

Thanks!

michal-z:
Hi,

PR would be much appreciated.

Thanks,
Michal

PR #332, opened 2023-07-06
rename for consistency

ssteinbach:
Sorry about that - that was from your code review but took me a second to hit both spots the rename needed to be.

Related to #328


ssteinbach:
Awesome thanks!

PR #336, opened 2023-07-08
Sample to Demonstrate zphysics DebugRenderer + Updates to DebugRenderer Itself

a-day-old-bagel:
If there is more procedure to go through before a sample may be approved, I can cherry pick out the parts that actually change zphysics itself (those changes are somewhat essential for the functioning of DebugRenderer).

Otherwise, the sample I've written provides a demonstration of the DebugRenderer, and I also plan to use it to demonstrate other parts of the currently-incomplete zphysics API as they are added. The fact that the sample isn't named something like `physics_debug_rendering_wgpu` is due to this idea of the sample demonstrating more than one part of the API in the future.

And the reasons that I didn't simply include this stuff in the existing `physics_test_wgpu` sample include the following:
1. I didn't author that sample, so I didn't want to completely overhaul it
2. `physics_test_wgpu` is a simple baseline "minimum viable product" of sorts to demonstrate the use of zphysics, and adding demonstrations of all the other stuff that is (or will be) in zphysics would run the risk of making it overcomplicated for learning purposes.

The sample currently consists of a scene with a number of moving visual elements, whose movement will appear mysterious to the user until they check the `Physics Debug Renderer - Draw Bodies` checkbox in the little imgui window that's visible in the sample. Once that box is checked, the sample will also render Jolt's debug-drawn objects to clarify what's going on.

image
image

Other parts of Jolt that I hope to demonstrate within this sample in the future, once they're implemented, include:
* The complete library and functionality of constraints
* The record/replay functionality of Jolt

As always, I'm happy to cherry pick, rebase, and otherwise wrangle this PR in any way if changes are desired.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -178,6 +178,7 @@ fn samplesCrossPlatform(b: *std.Build, options: Options) void {
     const layers_wgpu = @import("samples/layers_wgpu/build.zig");
     const gamepad_wgpu = @import("samples/gamepad_wgpu/build.zig");
     const physics_test_wgpu = @import("samples/physics_test_wgpu/build.zig");
+    const monolith = @import("samples/monolith/build.zig");
 
     install(b, minimal_gl.build(b, options), "minimal_gl");
     install(b, triangle_wgpu.build(b, options), "triangle_wgpu");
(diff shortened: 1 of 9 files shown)


michal-z:
Impressive work, thanks a lot!

Issue #175, opened 2022-12-23
zjobs: high cpu usage


Jack-Ji:
I'm trying to parallel frontend of my software rasterizer using zjobs. The default JobQueue occupied at least 80% cpu, even when I'm not scheduling any jobs. 

OS: windows 11 x64
zig: latest master

michal-z:
@garettbass is this known issue?

hazeycode:
zjobs has threads busy wait using sleep. You can configure the sleep time for your JobQueue by changing `idle_sleep_ns` form the default of 10 ns. Note this is probably below the accuracy error of the OS's scheduler. But by making it bigger you should see less CPU usage. By doing this you are trading "wake up" performance i.e. how long it can take a job to get picked up by a thread. You basically get to pick between wasting energy or increasing latency of jobs starting. You can get better by not using sleep and instead suspending and signalling threads (but it's kinda complicated to do cross-platform).
cyptocode write a nice blog post, showing how this looks at a high level using the async/await and standard event loop abstractions: https://dev.to/stein/async-cpu-bound-workers-in-zig-24b0

Jack-Ji:
I've thought of adjusting sleeping interval actually. However, I also   think the default configuration should works without obviously problem across platforms.

hazeycode:
> I've thought of adjusting sleeping interval actually. However, I also think the default configuration should works without obviously problem across platforms.

10 ns is really tiny! I'd guess that means basically that the thread is suspended and immediately woken by the OS scheduler.

Jack-Ji:
> > I've thought of adjusting sleeping interval actually. However, I also think the default configuration should works without obviously problem across platforms.
> 
> 
> 
> 10 ns is really tiny! I'd guess that means basically that the thread is suspended and immediately woken by the OS scheduler.

Yeah, I agree with you. May be we can change it to some more reasonable value? Any way, I'll try it later to see what happens.

hazeycode:
> > > I've thought of adjusting sleeping interval actually. However, I also think the default configuration should works without obviously problem across platforms.
> > 
> > 
> > 10 ns is really tiny! I'd guess that means basically that the thread is suspended and immediately woken by the OS scheduler.
> 
> Yeah, I agree with you. May be we can change it to some more reasonable value? Any way, I'll try it later to see what happens.

The problem is it's difficult to pick a number that is going to work well everywhere. You can provide a different value per target OS but you will also have to tune it per use case in your program.

[...]

Issue #130, opened 2022-10-08
zjobs: use Thread.Condition instead of sleep?


Avokadoen:
I'm just wondering if there is a reason `sleep` is used instead of `ResetEvent` for wait.
Has it been profiled to be faster?

I am willing to try to implement a ResetEvent based wait function if it has not been tested before. 

@garettbass

garettbass:
Thanks for the suggestion.  I am new to zig and was not aware of ResetEvent.  A PR is welcome.  I’m definitely no expert in this area, and just tried to create the simplest thing I could think of that might work.

garettbass:
Having read ResetEvent.zig, I think we could add a ResetEvent to each Slot (storage for pending jobs), we could call `reset` before executing a job, `set` after executing a job, and `waitForSet` instead of `sleep` when there is a `prereq`.  Does that sound right?

Avokadoen:
That sounds right. I double checked what I have used before and Conditions are probably what we want, not ResetEvent (my bad!)

A condition can be signaled and waited on. The tests in the file for Condition should be sufficient but I also used them in this file: https://github.com/Avokadoen/zig_vulkan/blob/main/src/modules/voxel_rt/brick/Worker.zig

Not currently at home to give more context. I'll get working on a prototype tomorrow if you don't want to try and implement it yourself 👍

Avokadoen:
And btw zjobs is even more impressive if this is one of your first Zig projects! 😍

michal-z:
> And btw zjobs is even more impressive if this is one of your first Zig projects! 😍

@garettbass also contributed `zpool` lib. Impressive work. There is also a chance we will get `zentity` some day :)

garettbass:
`zentity` is still in the works.  It is big, so I’m trying to break out smaller sections into useful libs, like `zjobs`.

Avokadoen:
I tried making a prototype for this. It should not be too hard but I see that it is more work than expected to refactor. Like all multithreading there seems to be a few ways to shoot yourself in the foot with this change. I will still look into it, but I'll not prioritize it currently because I have more pressing issues in my hobby project.

garettbass:
[...]

PR #338, opened 2023-07-09
zjobs: Remove use of thread sleep 

Avokadoen:
Related issues: #130, #175

cc @garettbass

### DIFF
--- a/libs/zjobs/src/zjobs.zig
+++ b/libs/zjobs/src/zjobs.zig
@@ -71,7 +71,7 @@ pub const JobId = enum(u32) {
 pub const QueueConfig = struct {
     max_jobs: u16 = 256,
     max_job_size: u16 = 64,
-    max_threads: u8 = 4,
+    max_threads: u8 = 32,
     idle_sleep_ns: u32 = 50,
 };
 
--- a/libs/zjobs/src/zjobs.zig
+++ b/libs/zjobs/src/zjobs.zig
@@ -68,6 +68,13 @@ pub const JobId = enum(u32) {
 
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 
+pub const QueueConfig = struct {
+    max_jobs: u16 = 256,
+    max_job_size: u16 = 64,
+    max_threads: u8 = 4,
+    idle_sleep_ns: u32 = 50,
+};
+
 /// Returns a struct that executes jobs on a pool of threads, which may be
 /// configured as follows:
 /// * `max_jobs` - the maximum number of jobs that can be waiting in the queue.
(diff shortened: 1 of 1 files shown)


Avokadoen:
There is also this WIP branch by @garettbass 
https://github.com/garettbass/zig-gamedev/tree/garett/zjobs-idle-queue 

It might be a better approach. I think changing the primitive to ResetEvent in the queue and then the current flow:

1. If jobid's entry index is not the current jobid then we don't check event (basically status quo)
2. If Entry is idle/empty then the event is set
3. If the Entry is executing jobid then it's reset

Avokadoen:
I tried adding `ResetEvent` to the Slot struct and then implement `wait(prereq: JobId)` with the following impl (based on isPending):

```zig
       /// Waits until the specified `prereq` is completed.
        pub fn wait(self: *Self, prereq: JobId) void {
            if (prereq == JobId.none) return;

            const _id = prereq.fields();
            assert(isLiveCycle(_id.cycle));

            const slot: *Slot = &self._slots[_id.index];
            const slot_cycle = slot.cycle.load(.Monotonic);

            if (slot_cycle == _id.cycle) {
                slot.idle_event.wait();
            }
        }
   ```
   
   We then just do the following in the top of executeJob in Slot: 
   
```zig
// while executing any observer can wait on the idle event
self.idle_event.reset();
defer self.idle_event.set();
```

The problem is that I hit deadlocks in the the "combine jobs respect prereq" test. 
It seems like a the wait is the cause of the deadlock

```zig
       /// Waits until the specified `prereq` is completed.
        pub fn wait(self: *Self, prereq: JobId) void {
            // ...
            const slot_cycle = slot.cycle.load(.Monotonic);
            
            // RC: slot_cycle might not be the active slot_cycle when we compare and when we call wait
            if (slot_cycle == _id.cycle) { 
                slot.idle_event.wait();
            }
        }
```

I might be able to resolve the deadlock by adding a mutex to the slots (which means a condition might be a better fit) :thinking:

Avokadoen:
I have a working version with the mutex + condition wrapping any write/read of the slot with mutex. I'll profile vs main branch to see if there are gains from this and if cpu active improves. I can't run the throughput test as it seems to be crashing on my   i7-6500U 

I also want to make it so the enqueue only wakes a single thread instead of all threads

Avokadoen:
Blocking individual threads have proven to be a challenge :sweat_smile: 
I think it should be fine to open a issue on that instead of trying to solve that in this PR.

I am currently traveling and I have issues with my laptop so this PR won't be marked as ready until I come back on Sunday and have the opportunity to benchmark this patch.

Avokadoen:
@garettbass I'm marking this as ready for review. Overall it seems like this actually might cause minor regression in some scenarios. I ran throughput which is aprox 10ms slower on i7 6700k (~310ms to ~320ms). I also have a [game of life example in ecez](https://github.com/Avokadoen/ecez/blob/main/examples/game-of-life/main.zig) which does not show either regression or improvements although there is no long running jobs in either. I think we would see less CPU active in longer running jobs, especially if we could get threads to individually wake up when scheduling new work.

Avokadoen:
For reference this is the results of the game of life example running for ~34s where there is some improvement with the PR patches:

- 34.06s vs 38.89s total time spent on waiting
- min cpu util of 68.75% vs 79.17%

image

garettbass:
I did some work on waking threads using condition variables last year, but never completed the refactor.  I have no objection to these changes and likely will not be involved in further development of this library in the near future.

Avokadoen:
Thanks for the heads up and all the amazing work you have done so far @garettbass! ❤️

PR #339, opened 2023-07-11
zphysics: Added Decorated Shapes

a-day-old-bagel:
Let me know if this isn't the way you'd put this part of the API together. I can redo this any way you'd like.

I just threw something in there quickly because I needed it, and I've tested it by using it in my own project.

I haven't written these into the actual tests yet though.

I combined each of Jolt's three existing inheritors of the DecoratedShape base class into the same type in the Zig API, because including the complexity of that two-deep inheritance seemed unnecessary. Again, let me know if you disagree.


michal-z:
Thanks, looks good!

> I combined each of Jolt's three existing inheritors of the DecoratedShape base class into the same type in the Zig API, because including the complexity of that two-deep inheritance seemed unnecessary. Again, let me know if you disagree.

I would the same.

PR #341, opened 2023-07-12
zsdl: Implement texture locking/unlocking

GeffDev:


libs/zsdl/src/sdl2.zig
```diff
@@ -383,6 +383,11 @@ pub const Access = enum(i32) {
 };
 
 pub const Texture = opaque {
+    pub const PixelData = struct {
+        pixels: [*]u8,
+        pitch: i32,
```
hazeycode, line 388: `pitch` should be a `u32` in the public API, the `i32` returned from SDL can be `@bitCast` if the `LockTexture` call was successful.

Also, we could consider calculating the length of `pixels` and using a slice here instead of a pointer. This would add some safety to the API.

libs/zsdl/src/sdl2.zig
```diff
@@ -394,6 +399,25 @@ pub const Texture = opaque {
...
+        if (SDL_LockTexture(texture, rect, &pixels, &pitch) != 0) {
+            return makeError();
+        }
+
+        return PixelData{
+            .pixels = @ptrCast(pixels),
+            .pitch = pitch,
+        };
+    }
+    extern fn SDL_LockTexture(texture: ?*Texture, rect: ?*Rect, pixels: ?**anyopaque, pitch: ?*c_int) c_int;
```
hazeycode, line 415: I think the `texture` fn args here should not be optional.

hazeycode (changes requested):

libs/zsdl/src/sdl2.zig
GeffDev: Fixed this up.

hazeycode (approved):

libs/zsdl/src/sdl2.zig
GeffDev: Cool, pitch is now `u32`.

> Aside: we could consider calculating the length of pixels and using a slice here instead of a pointer. This would add some safety to the API. But at the cost of "thickening" otherwise thin bindings, which may not be desirable.

Would a `len` field (w * h) work then? Or should the user just use `query` for that?

hazeycode: It would be something like (I haven't tested this)
```zig
const width = rect.w * pixel_format.BytesPerPixel;
assert(pitch >= width)
const len = rect.h * pitch - (pitch - width);
const pixels = pixels_ptr[0..len];
```
Note: This would only prevent the user from overrunning the end of the locked region; it wouldn't prevent writing outside the locked region if `pitch` > `width`

PR #351, opened 2023-07-22
zsdl: Order by API categories per SDL docs

hazeycode:
Just some housekeeping, ordering bindings by API categories per the docs: https://wiki.libsdl.org/SDL2/APIByCategory

This will make it easier to diff SDL2 and SDL3 bindings going forward.

Note that API by Category docs for both SDL2 and SDL3 are missing some items currently. So there will be further changes here to reflect changes to upstream docs.

Upstream SDL wiki PRs:
- https://github.com/libsdl-org/sdlwiki/pull/396
- https://github.com/libsdl-org/sdlwiki/pull/397

### DIFF
--- a/libs/zsdl/src/sdl2.zig
+++ b/libs/zsdl/src/sdl2.zig
@@ -1,9 +1,10 @@
 const builtin = @import("builtin");
 const std = @import("std");
 const assert = std.debug.assert;
+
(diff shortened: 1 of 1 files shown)


michal-z:
Nice, thanks!

PR #354, opened 2023-07-24
Add `comboFromEnum` function

### LINKED ISSUE #353: [zgui] combo from enum?
I wrote this helper, to turn an enum into a combo directly:

```zig
pub fn combo_from_enum(
    label: [:0]const u8,
    current_item: anytype
) bool 
{
    const item_names = comptime lbl: {
        const item_type = @typeInfo(@TypeOf(current_item.*));
        switch (item_type) {
            .Enum => |e| {
                comptime var str: [:0]const u8 = "";

                inline for (e.fields) |f| {
                    str = str ++ f.name ++ "\x00";
                }
                break :lbl str;
            },
            else => {
                @compileError(
                    "Error: current_item must be a pointer-to-an-enum, not a "
                    ++ @TypeOf(current_item)
                );
            }
        }
    };

    var item = @intCast(i32, @intFromEnum(current_item.*));

    const result = combo(
        label,
        .{
            .items_separated_by_zeros = item_names,
            .current_item = &item,
        }
    );

    current_item.* = @enumFromInt(@TypeOf(current_item.*), item);

    return result;
}
```

If I throw that into a PR, is that something this library would want?  Or do you want to stick to the more direct imgui wrapper?

Thanks again for the great library!

ssteinbach:
Closes #353 

* wrapper around `combo` that uses comptime reflection to turn an enum's possibilities into options for a combo
* allows for making dropdowns that drive enums really easily, example:

```zig
    const FoodKinds = enum(i8) {
        banana = 0,
        taco,
        spaghetti,
    };

    var myEnum : FoodKinds = .banana;

    _ = zgui.comboFromEnum( "Food", &myEnum);
```

Makes this combobox:
Screenshot 2023-07-24 at 10 32 50 AM

### DIFF
--- a/libs/zgui/src/gui.zig
+++ b/libs/zgui/src/gui.zig
@@ -1487,7 +1487,7 @@ pub fn comboFromEnum(
         }
     };
 
-    var item = @intCast(@intFromEnum(current_item.*));
+    var item:i32 = @intCast(@intFromEnum(current_item.*));
 
     const result = combo(
         label,
@@ -1497,7 +1497,7 @@ pub fn comboFromEnum(
         }
     );
 
-    current_item.* = @enumFromInt(@TypeOf(current_item.*), item);
+    current_item.* = @enumFromInt(item);
 
     return result;
 }


libs/zgui/src/gui.zig
```diff
@@ -1450,6 +1450,50 @@ pub fn combo(label: [:0]const u8, args: Combo) bool {
...
+            else => {
+                @compileError(
+                    "Error: current_item must be a pointer-to-an-enum, not a "
+                    ++ @TypeOf(current_item)
+                );
+            }
+        }
+    };
+
+    var item = @intCast(i32, @intFromEnum(current_item.*));
```
michal-z, line 1483: Looks like you are using older Zig compiler, casts take only one arg now. Below should work:
`var item: i32 = @intCast(@intFromEnum(current_item.*));`

michal-z (changes requested):
Thanks! Found one issue.

libs/zgui/src/gui.zig
```diff
@@ -1450,6 +1450,50 @@ pub fn combo(label: [:0]const u8, args: Combo) bool {
...
     );
 }
+/// creates a combo box directly from a pointer to an enum value using zig's
+/// comptime mechanics to infer the items for the list at compile time
+pub fn comboFromEnum(
+    label: [:0]const u8,
+    /// must be a pointer to an enum value (var my_enum: *FoodKinds = .Banana)
+    /// that is backed by some kind of integer that can safely cast into an
+    /// i32 (the underlying imgui restriction)
+    current_item: anytype
```
xdBronch, line 1460: should add a comma after this last argument otherwise zig fmt will do some weird stuff

libs/zgui/src/gui.zig
ssteinbach: Ah, yup.  We're locked on our zig version, too much churn to follow head.  I've downloaded the latest and switched to the new syntax. Thanks!

libs/zgui/src/gui.zig
ssteinbach: Cool, good to know.  We don't use `zig fmt` so this wasn't on my radar.

ssteinbach:
Pushed post review notes, thanks for the feedback, @michal-z and @xdBronch!

ssteinbach:
(also rebased onto latest)

PR #356, opened 2023-07-25
Some small changes to zflecs.

Srekel:


Srekel:
Yeah, I thought there was a difference between tags and zero-size components but apparently that's simply what it is :) I looked into various git magics yesterday to see if I could redo the PR without polluting the timeline but... I think I prefer doing that to accidentally causing a rift in time and space to open and letting the Chaos gods come in.

Srekel:
I simply did a reverse commit. :)

Avokadoen:
Optionally the merge can be done with a squash to remove a invalid commit

PR #362, opened 2023-07-26
Added function that makes Mat from [16]f32

d0miKuras:


michal-z:
Thanks! Can you also add this function to the list at the top of the file (docs)?

PR #363, opened 2023-07-30
Add option to zgui to use imgui out of source

iacore:
For example, if I want to use other backends, it's better to link in imgui myself


michal-z:
Yes, makes sense. Thanks for the PR!

PR #365, opened 2023-07-30
zgui: Add option to not use implot

iacore:
I'd say it's better to not merge this for two weeks. Who knows what other change is to come?


hazeycode:
Hey @iacore what's the intention here? Does this fix a particular issue for you?

iacore:
The issue back then was that building implot caused some trouble, so I had to left it out of the build.

hazeycode:
> The issue back then was that building implot caused some trouble, so I had to left it out of the build.

I guess there could be symbol clashes depending on how you are compiling things.

The changes look good to me. Is this ready to merge?

iacore:
> Is this ready to merge?

yes

PR #366, opened 2023-08-01
zgpu: Expose present mode on context creation

tdm42:
Having vsync off can be desirable sometimes, but it's currently forced on

### DIFF
--- a/libs/zgpu/src/zgpu.zig
+++ b/libs/zgpu/src/zgpu.zig
@@ -51,7 +51,7 @@ pub const GraphicsContext = struct {
         } = .{},
     } = .{},
 
-    pub fn create(allocator: std.mem.Allocator, window: *zglfw.Window) !*GraphicsContext {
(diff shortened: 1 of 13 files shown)


michal-z:
Yes this is very useful, thanks for the PR.

I would like to do it in a more extensible way because we may want to add more options in the future.

```zig
pub const GraphicsContextOptions = struct {
    present_mode: wgpu.PresentMode = .fifo,
};

pub const GraphicsContext = struct {
    pub fn create(allocator: std.mem.Allocator, window: *zglfw.Window, options: GraphicsContextOptions) !*GraphicsContext {
};
```

Usage:
```zig
GraphicsContext.create(allocator, window, .{ .present_mode = .fifo });
```

tdm42:
Alright, updated it to match the above example (although I didn't explicitly specify the fifo render mode in the examples, since that's the default value)

michal-z:
Thanks!

PR #368, opened 2023-08-04
ztracy: use new LazyPath enum in build script

Avokadoen:
this fixes build issues


Avokadoen:
Sorry for the noise, I used old code before.

PR #375, opened 2023-08-13
zphysics: Added BodyInterface.GetMotionType

Srekel:
Let me know if you want me to keep or revert the fmt comments. I added them so that I don't have to revert the formatting changes whenever I save the file (I use format-on-save)


Srekel:
Ah I mean I can just remove the fmt comments, I assumed the formatting was that way intentionally :) It doesn't affect much, primarily the alignment of the comments.

PR #376, opened 2023-08-13
zsdl: Add Version, SDL_VERSION, and getVersion

scbafk:
Adds Version, SDL_VERSION, and getVersion to zsdl.

### DIFF
--- a/libs/zsdl/src/sdl2.zig
+++ b/libs/zsdl/src/sdl2.zig
@@ -103,6 +103,28 @@ pub fn makeError() error{SdlError} {
 // Querying SDL Version
 //
 //--------------------------------------------------------------------------------------------------
+pub const Version = extern struct {
+    major: u8,
+    minor: u8,
+    patch: u8,
+};
+
+// Compile-time Version
+pub inline fn SDL_VERSION(version: *Version) void {
+    version.major = SDL_MAJOR_VERSION;
+    version.minor = SDL_MINOR_VERSION;
+    version.patch = SDL_PATCHLEVEL;
+}
+
+pub const SDL_MAJOR_VERSION: u8 = 2;
+pub const SDL_MINOR_VERSION: u8 = 24;
+pub const SDL_PATCHLEVEL: u8 = 1;
+
+// Runtime Version
+pub fn getVersion(version: *Version) void {
+    SDL_GetVersion(version);
+}
+extern fn SDL_GetVersion(version: *Version) void;
 
 //--------------------------------------------------------------------------------------------------
 //


libs/zsdl/src/sdl2.zig
```diff
@@ -103,6 +103,28 @@ pub fn makeError() error{SdlError} {
...
+// Compile-time Version
+pub inline fn SDL_VERSION(version: *Version) void {
+    version.major = SDL_MAJOR_VERSION;
+    version.minor = SDL_MINOR_VERSION;
+    version.patch = SDL_PATCHLEVEL;
+}
+
+pub const SDL_MAJOR_VERSION: u8 = 2;
+pub const SDL_MINOR_VERSION: u8 = 24;
+pub const SDL_PATCHLEVEL: u8 = 1;
```
hazeycode, line 121: This could be simplified a bit:
```zig
/// Compiled SDL version
pub const VERSION = Version{
    .major = 2,
    .minor = 24,
    .patch = 1,
};
```

libs/zsdl/src/sdl2.zig
```diff
@@ -103,6 +103,28 @@ pub fn makeError() error{SdlError} {
...
+
+// Runtime Version
```
hazeycode, line 123: I think we should stick to SDL terminology and use doc-comments (`///`) when describing types and fns:
```zig
/// Returns the linked SDL version
```

hazeycode (changes requested):

libs/zsdl/src/sdl2.zig
scbafk: Good call. I did a straight port, but this is better.

hazeycode (approved):
Nice!

libs/zsdl/src/sdl2.zig
```diff
@@ -103,6 +103,24 @@ pub fn makeError() error{SdlError} {
...
+
+/// Compiled SDL version
+pub const VERSION = Version{
+    .major = 2,
+    .minor = 24,
+    .patch = 1,
+};
+
+/// Returns the linked SDL version
+pub fn getVersion(version: *Version) void {
```
hazeycode, line 120: We could be more consistent with other bindings here if we don't pass a ptr to Version and instead construct and return from the fn. Generally we only stick to the C paradigm when it's valid to pass null, which is not the case here.
The signature would be:
`pub fn getVersion() Version`
What do you think?

scbafk: I like it :+1:

michal-z:
Thanks!

PR #377, opened 2023-08-13
zsdl: Add SDL_GetWindowWMInfo

scbafk:
First pass at wrapping `SDL_GetWindowWMInfo`. Only have access to windows and linux platforms so only "enabled" it for those two. 

I don't know what platforms zsdl supports so I'll open this as a draft for now and we can determine what is a good fit.

I also could not determine some types in the Android and Vivante structs so I left those as the default union types. 

Note: `SDL_GetWindowWMInfo` requires using `SDL_VERSION` on info struct before calling.


hazeycode:
I'd say don't worry about supported platforms. It would be nice to support every platform that SDL supports; but just implementing those we use and leaving TODOs is fine for now I think.

libs/zsdl/src/sdl2.zig
```diff
@@ -1056,6 +1056,101 @@ pub const Surface = opaque {
...
+    info: extern union {
+        win: SysWMInfo_win,
+        x11: SysWMInfo_x11,
+        winrt: SysWMInfo_winrt,
+        dfb: SysWMInfo_dfb,
+        cocoa: SysWMInfo_cocoa,
+        uikit: SysWMInfo_uikit,
+        wl: SysWMInfo_wayland,
+        // Android -- Uncertain on struct definition
+        // Vivante -- Uncertain on struct definition
```
hazeycode, line 1147: I think `EGLSurface`, `EGLNativeDisplayType` and `EGLNativeWindowType` should be opaque pointers; at least they are in Angle.

hazeycode:
For Android, I think `surface` should be an opaque pointer (see https://github.com/KhronosGroup/EGL-Registry/blob/main/api/1.2/EGL/egl.h); `window` too.

I don't know about Vivante. I think it probably uses the normal EGL headers but we could just leave a TODO here for now.

I do think we are probably safe assuming opaque pointers per the type definitions from the EGL registry. For instance, in Angle, for emcripten and symbian, EGL types are defined differently but they are always pointer sized, which I would expect to be the case everywhere.

libs/zsdl/src/sdl2.zig
```diff
@@ -1056,6 +1056,101 @@ pub const Surface = opaque {
...
+        // MIR -- Did not include, SDL unsupported and recommended to drop after 2.1
+        android: [64]u8,
+        vivante: [64]u8,
+        dummy: [64]u8,
```
hazeycode, line 1151: We can do a comptime assert here and delete the `dummy` variant
```zig
comptime {
    std.debug.assert(@sizeOf(SysWMInfo) == 64);
}
```

scbafk:
> For the EGL types, see https://github.com/KhronosGroup/EGL-Registry/blob/main/api/EGL/egl.h and https://github.com/KhronosGroup/EGL-Registry/blob/main/api/EGL/eglplatform.h
> 
> opaque pointers should be fine for `EGLSurface` and others, since they appear to always be pointer sized types even where defined differently for different platforms

Great, thanks for figuring this out.

libs/zsdl/src/sdl2.zig
```diff
@@ -1056,6 +1056,117 @@ pub const Surface = opaque {
...
+    haiku,
+    kmsdrm,
+    riscos,
+};
+
+/// Get driver-specific information about a window.
+///
+/// The caller must initialize the `info` structure's version by using
+/// `VERSION(&info.version)`, and then this function will fill in the rest
+/// of the structure with information about the given window.
```
hazeycode, line 1081: This comment is out of date. `VERSION` is now a constant.

libs/zsdl/src/sdl2.zig
```diff
@@ -1056,6 +1056,117 @@ pub const Surface = opaque {
...
+/// of the `info` struct is valid, or false if the information
+/// could not be retrieved
+///
+/// ZSDL Note: Only implemented for SysWMType.windows & SysWMType.x11
+pub fn getWindowWMInfo(window: *Window, info: *SysWMInfo) bool {
+    const result: bool = SDL_GetWindowWMInfo(window, info);
+    return switch (info.subsystem) {
+        .windows, .x11 => result,
+        else => false,
+    };
```
hazeycode, line 1093: Why don't we just return the result of `SDL_GetWindowWMInfo` here? Why are windows and x11 special?

libs/zsdl/src/sdl2.zig
```diff
@@ -1056,6 +1056,117 @@ pub const Surface = opaque {
...
+}
+extern fn SDL_GetWindowWMInfo(window: *Window, info: *SysWMInfo) bool;
```
hazeycode, line 1095: The return type should be `Bool` rather than `bool`

libs/zsdl/src/sdl2.zig
```diff
@@ -1056,6 +1056,117 @@ pub const Surface = opaque {
...
+    android: SysWMInfo_android,
+    vivante: SysWMInfo_vivante,
+    // MIR -- SDL unsupported and recommended to drop after 2.1
+
+    // This union has a "soft" maximum size of 64 bytes
+    // SDL denotes this with a dummy value of [64]u8
+    // We can use a comptime assert to make this a hard requirement
+    comptime {
+        assert(@sizeOf(SysWMInfo_info) <= 64);
+    }
```
hazeycode, line 1162: `SysWMInfo_info` should be exactly 64 bytes, see the comment here: https://github.com/native-toolkit/libsdl/blob/fa51fcb56b152ba6419b6decfef5dad48b075ac2/include/SDL_syswm.h#L291C9-L292C71

Actually, we may need to have a `dummy` variant to ensure that this is the case if other variants are are smaller, sorry for any confusion here.  Either way it is still useful to have a comprise assert.

libs/zsdl/src/sdl2.zig
```diff
@@ -1056,6 +1056,117 @@ pub const Surface = opaque {
...
+};
+
+pub const SysWMInfo = extern struct {
+    version: Version,
+    subsystem: SysWMType,
+    info: SysWMInfo_info,
```
hazeycode, line 1168: Stylistic nitpick: I think it would be nicer to define `info` inline, as an anonymous union, similar to how it is defined in SDL. To get the size of the anonymous union use `@sizeOf(@This())` in the comptime assert

hazeycode (changes requested):

libs/zsdl/src/sdl2.zig
scbafk: All good, makes sense

hazeycode (approved):
Thanks for working on this!

PR #378, opened 2023-08-15
Fixed zflecs README.md instructions

Pyrolistical:
Added
 * missing exe.addModule
 * missing structs
 * expected result

### DIFF
--- a/libs/zflecs/README.md
+++ b/libs/zflecs/README.md
@@ -17,7 +17,6 @@ pub fn build(b: *std.Build) void {
 
     const zflecs_pkg = zflecs.package(b, target, optimize, .{});
 
-    exe.addModule("zflecs", zflecs_pkg.zflecs);
     zflecs_pkg.link(exe);
 }
 ```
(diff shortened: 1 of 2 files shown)


Pyrolistical:
Also, I think `exe.addModule("zflecs", zflecs_pkg.zflecs);` would be more clear as `exe.addModule("zflecs", zflecs_pkg.module);` if we rename: https://github.com/michal-z/zig-gamedev/blob/d48b1abbd8facce124d2f15b7738272718538ee8/libs/zflecs/build.zig#L4

michal-z:
Thanks for the PR.

`exe.addModule()` should be added to `Package.link()` method. User only needs to call `zflecs.package()` to build the package and then `zflecs_pkg.link()` to link the lib with their app. This is our rule and it is true for all libs.

Rationale for the naming:
`Package.zflecs` is a main module of the package. In the future package can get more modules, for example: `Package.zflecs_options`.

Again, we use this naming for all our libs, for example:

```zig
pub const Package = struct {
    zgui: *std.Build.Module,
    zgui_options: *std.Build.Module,
    zgui_c_cpp: *std.Build.CompileStep,
    ...
};
```

Pyrolistical:
Thanks, I've moved the `exe.addModule` into `build.zig`.

PR #380, opened 2023-08-19
Fix breaking zig change: deprecated float NaN constants

kamidev:
This commit has breaking changes: https://github.com/ziglang/zig/commit/387b0ac4f1c54cb2f83792299aa628a316e17d88


michal-z:
Thanks!

PR #384, opened 2023-08-20
zsdl: SDL3 bindings WIP

hazeycode:
Subject to further change (breaking changes are still being made upstream).

Except for API additions to `Color` that I don't think should be in sdl2.zig, this is complete enough that sdl2.zig and sdl3.zig are diffable.

I will followup with binaries, updated build.zig and make these bindings optionally available to zsdl users in a future PR.


michal-z:
Awesome, thanks!

PR #385, opened 2023-08-20
zsdl: Add basic compile test & fixes

hazeycode:
* Added basic compilation test
* Fixed a number of compile errors
* Fixedup some types
* Also deleted `Color` API additions that do not exist in SDL. Those are better implemented and provided outside this bindings lib @Jack-Ji


libs/zsdl/src/sdl2.zig
```diff
@@ -759,26 +743,22 @@ pub const Renderer = opaque {
...
 
     pub fn readPixels(
-        r: *Renderer,
+        renderer: *const Renderer,
         rect: ?*const Rect,
-        format: ?PixelFormatEnum,
+        format: PixelFormatEnum,
         pixels: [*]u8,
-        pitch: u32,
+        pitch: u16,
```
Jack-Ji, line 750: Should use either c_int or i32

libs/zsdl/src/sdl3.zig
```diff
@@ -688,46 +676,42 @@ pub const Renderer = opaque {
...
 
     pub fn readPixels(
-        r: *Renderer,
+        renderer: *const Renderer,
         _rect: ?*const Rect,
-        format: ?PixelFormatEnum,
+        format: PixelFormatEnum,
         pixels: [*]u8,
-        pitch: u32,
+        pitch: u16,
```
Jack-Ji, line 703: Should use either c_int or i32

Jack-Ji (changes requested):

PR #387, opened 2023-08-21
Add sample minimal_zgpu_zgui

MyNameIsTrez:
I am deliberately letting the program crash at the moment, rather than using the commented out line in this sample's build.zig, since the commented out line wasn't necessary in the other samples. In other samples `exe_options.addOption([]const u8, "content_dir", content_dir);` "just works".

If someone could verify that the program also crashes for them, that'd be great. A tip on why the crash occurs/how to fix the crash would be even better, of course! :)

Any stylistic feedback that doesn't add more than one or two lines of code is also welcome.


MyNameIsTrez:
@michal-z Uhh... you weren't supposed to merge it yet, cause as I said it is currently deliberately crashing, waiting for feedback on how to *properly* fix it, haha. Or maybe you wanted to merge it already and fix it yourself later?

michal-z:
Sorry, I was too fast :)

I did fix it with the latest commit.

Please take a look what was missing.

MyNameIsTrez:
@michal-z Awesome, thank you! :)

PR #391, opened 2023-08-24
zsdl: Add pushEvent

scbafk:


libs/zsdl/src/sdl2.zig
```diff
@@ -1251,6 +1251,10 @@ pub fn pollEvent(event: ?*Event) bool {
 }
 extern fn SDL_PollEvent(event: ?*Event) i32;
 
+/// pub fn pushEvent(event: ?*Event) i32 {
+pub const pushEvent = SDL_PushEvent;
+extern fn SDL_PushEvent(event: ?*Event) i32;
```
hazeycode, line 1256: * the `event` arg should be of type `*Event`, as it's not valid to pass null to this API
* Because `SDL_PushEvent` only returns a result code we should check for this and change the return type of `pushEvent` to `Error!void`. See `showSimpleMessageBox` for an example of this.

hazeycode (changes requested):

libs/zsdl/src/sdl2.zig
michal-z: I think we should use `Error!bool` (`true` == added to the queue, `false` == filtered). What do you think?

scbafk: Another option is `Error!bool`, true for event added and false for event filtered.

scbafk: @michal-z That sounds good to  me.

hazeycode: Good catch. Sounds good to me too 👍

PR #395, opened 2023-08-27
minimal glfw opengl sample

hazeycode:
* zglfw: added `windowHint`, `windowHintString` and `windowHintTyped`
* zglfw: deprecated `WindowHint.set` (see comment)
* zglfw: defined `ContextRobustness`, `ReleaseBehaviour` and `ContextCreationApi`
* zglfw: added `getProcAddress`
* added minimal_glfw_gl sample
* renamed minimal_gl to minimal_sdl_gl


michal-z:
Great, thanks!

Just one minor thing, I would prefer to declare `GlProc` as:
`pub const GlProc = *const anyopaque`
and then use it as: `?GlProc`

I think this is more readable.

hazeycode:
@michal-z I added some more to this PR. Notably, `WindowHint.set` is deprecated because it does not allow user to set string hints. Users should instead use `windowHint` or `windowHintString`. Additionally, `windowHintTyped` can be used in most use cases. I have updated minimal_glfw_gl to use the typed variant but other uses are using the deprecated method.

PR #396, opened 2023-08-30
build system exploration

hazeycode:
Changes:
* Separation of samples, experiments, tests and benchmarks build routines and options (*not sure about this, just experimenting*)
* Simplified samples build code

TODO:
* Refactor and move `makeDxc` from each sample to `zd3d12` lib
* Move common shader compilation steps from each sample to `common` lib
* Consider making samples buildable standalone
* Consider reverting samples package defs back to top-level build.zig

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@
 zig-cache
 zig-out
 backup
+
+# Ignore special system paths
+*.DS_Store
--- a/build.zig
+++ b/build.zig
@@ -5,263 +5,92 @@ pub const min_zig_version = std.SemanticVersion{ .major = 0, .minor = 12, .patch
 
 pub fn build(b: *std.Build) void {
     //
-    // Options and system checks
+    // Environment checks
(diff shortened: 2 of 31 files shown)


michal-z:
> Samples and experiments configure lib packages for themselves

Please note that this is what we had in the past and this is not ideal because libs are being built multiple times. For example, if two samples use `zphysics` Zig will build `zphysics` twice.

hazeycode:
The problem with the current solution is that samples cannot individually configure packages, they all have the same options, this was the main motivation here because we cannot have zgui samples with different backends, for example.
I guess we could have multiple variants of lib packages, configured with different options, shared by samples?

hazeycode:
The problem I was trying to get around was having different package configurations per sample.

Thinking about it now, it would be better to configure packages via `samples/build.zig::Options`. To change the zgui backend for example.

I will change back to sharing packages across samples.

PR #397, opened 2023-09-01
update readme for optimization flags

jeremy-coleman:
i had to search around for this since the flags keep changing, and wasn't sure what the defaults were in the 500 lines of code of the build file. i would expect this repo and samples to be most game developer's first try with zig, so i think having the optimization options at least mentioned could prevent some people from having a skewed first impression of perf capability in a real world game/app. anyway, hope its helpful


michal-z:
Thanks!

PR #398, opened 2023-09-02
build.zig: simplify samples installation

hazeycode:


hazeycode:
Prevents samples from having custom build or install arguments with not much benefit

PR #403, opened 2023-09-09
w32 additions

nickrolfe:
- Add [GetWindowLongPtrA](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getwindowlongptra) and [SetWindowLongPtrA](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowlongptra).
- Add a few more [WM_* window message](https://learn.microsoft.com/en-us/windows/win32/winmsg/window-notifications) definitions.


michal-z:
Nice thanks!

`callconv(WINAPI)` is missing, please add it.

PR #404, opened 2023-09-12
Enable having zmath as a dependency

Game4all:
This aims to be a first step / PoC in enabling use of the libs in this repo as individual dependencies.
This justs exposes the zmath package so that it can added as a dependency in a `build.zig` file with:
```zig
    b.dependency("zmath", .{}); // .{ .enable_cross_platform_determinism = false } or .{} which defaults to .enable_cross_platform_determinism = true
    exe.addModule("zmath", zmath.module("zmath"));
```
If this is deemed fine, this can be applied to others libs later on.

No other build file has been touched (namely samples), guess it would be better to do it in one fell swoop once all libs are available as dependencies.

### DIFF
--- a/libs/zmath/build.zig
+++ b/libs/zmath/build.zig
@@ -32,7 +32,7 @@ pub fn package(
 
     const zmath_options = step.createModule();
 
-    const zmath = b.createModule(.{
+    const zmath = b.addModule("zmath", .{
         .source_file = .{ .path = thisDir() ++ "/src/main.zig" },
         .dependencies = &.{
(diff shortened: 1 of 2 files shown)


hazeycode (changes requested):
Hey, thanks for working on this!

Please also add a build.zig.zon in the zmath directory as has been done [for zopengl](https://github.com/michal-z/zig-gamedev/pull/424) and [for zglfw](https://github.com/michal-z/zig-gamedev/pull/428)

hazeycode (approved):

PR #417, opened 2023-10-03
zgui: Add glfw/opengl3 backend

kanewallmann:
Implemented in a similar way as #227. Using the exported GL bindings from zopengl instead of relying on imgui's own loader.

Tested to work on Linux.


hazeycode:
Have you got an example program using this?

zopengl's C exports still aren't working for me: https://github.com/kanewallmann/zig-gamedev/pull/1

Tested on macOS 13

hazeycode (approved):
https://github.com/kanewallmann/zig-gamedev/pull/1 is working for me now ~~(with caveat)~~

Nice work!

hazeycode (changes requested):
We should at least this change to backend_glfw_gl before merging: https://github.com/kanewallmann/zig-gamedev/pull/1/commits/257d16724c2f0fde44fb6e7e8b62e6fd5e6ff281#diff-5ce0ed694c553249b6fc4dda420b9991c73a2c3a3cf1658fc87e9856a3cd115a, or the whole sample.

kanewallmann:
Merged in your sample + the fix to make it consistent with wgpu thanks!

hazeycode (approved):

PR #418, opened 2023-10-03
Enable skipped zmath tests

kamidev:
Enable 3 skipped tests that now work on macOS.

There are still 4 skipped tests, they fail after calling 'math.inf'.


michal-z:
Thanks!

PR #421, opened 2023-10-10
zflecs: Added zflecs_init and zflecs_fini.

Srekel:
Hi, this PR is very open to feedback but I think the core of it is reasonable. It moves a few lines out of zflecs.init and zflecs.fini to two new functions, zflecs.zflecs_init and zflecs.zflecs_fini.

The purpose/result is twofold:
- Take a step towards zflecs supporting more than one world by moving away things that doesn't strictly have to do with world lifetimes from the world lifetime functions. 
- Proper setup of the abort function. Previously, in the order it was done, meant that it would in fact get overridden by flecs to the default handler and was therefore in fact not used at all. Now if flecs asserts while debugging, we will end up at the `@breakpoint` line.

Here's what my game code looks like:
```

    ecs.zflecs_init();
    defer ecs.zflecs_fini();
    var ecsu_world = ecsu.World.init();
    defer ecsu_world.deinit();
```

### DIFF
--- a/libs/zflecs/src/zflecs.zig
+++ b/libs/zflecs/src/zflecs.zig
@@ -976,22 +976,35 @@ fn flecs_abort() callconv(.C) noreturn {
 // Creation & Deletion
 //
 //--------------------------------------------------------------------------------------------------
-pub fn init() *world_t {
+pub fn zflecs_init() void {
+    os_init();
+    var api = os_get_api();
+
     if (builtin.os.tag == .windows) {
-        os.ecs_os_api.abort_ = flecs_abort;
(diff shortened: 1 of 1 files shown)


hazeycode:
Hey @Srekel are you still wanting to merge this?

I'm not very familiar with flecs but this refactor looks good to me. Although I wonder if the `init` and `fini` should be moved to `World.init` and `World.fini` and then have `zflecs_init` and `zflecs_fini` take the names `init` and `fini`, or am I reading this wrong?

Srekel:
I'll need to get back to you on this a bit later once I've double-checked the changes :)

PR #423, opened 2023-10-11
Add support for 2-bit depth textures

gmodarelli:
Hello 👋 
We've started working on UI for Tides of Revival and we encountered this missing format from wincoded.


michal-z:
Thanks!

PR #425, opened 2023-10-12
zphysics: Implemented Shape.getCenterOfMass

a-day-old-bagel:
Which is useful for getting the translation from Jolt's calculated center of mass frame back to the frame you used when creating the shape, if those are different. In the case of convex hulls, for example, they're often different.


michal-z:
Thanks!

PR #427, opened 2023-10-16
zd3d12: Allow window resizing

gmodarelli:
This PR allows apps that use zd3d12 to resize their window.
I also introduced a new struct for `SwapchainBuffer`s because before we were assuming that swapchain buffers were always allocated at the beginning of the `rtv_heap`. But that's no longer the case, so now we also store the `CPU_DESCRIPTOR_HANDLE` along side the `ResourceHandle`.


michal-z:
Looks good, thanks!

PR #431, opened 2023-10-24
child_process.ChildProcess.exec has been renamed to run, updating the…

Joen-UnLogick:
I just tried building with latest zig version and it failed. 

child_process.ChildProcess.exec has been renamed to run, updating the build script to match.


michal-z:
Thanks!

PR #433, opened 2023-10-28
zphysics: Constraint implementation began - FixedConstraint tested

a-day-old-bagel:
This lays out the beginning for the implementations of each constraint type, following the model used for the Shape types.

It works for FixedConstraint, so I'm pretty sure this is going to work going forward.

I was kind of rushed this time writing it, so you should probably double-check for any mistakes.

As always, let me know if you'd like this done differently, and I can make changes.


michal-z:
Thanks! Looks good, will merge today (sorry for late reply).

a-day-old-bagel:
> Thanks! Looks good, will merge today (sorry for late reply).

Oops, I didn't realize this wasn't merged yet, and I just pushed a change to my fork to fix the new compiler error that checks for usage of `var` where `const` could have been used.

PR #458, opened 2024-01-07
zgui: Expose slices into DrawList data

### LINKED ISSUE #372: zgui: Expose slices into DrawList data
It would be useful to have functions that return slices into the various buffers of the draw lists. Current design makes it a bit verbose, especially since the `len` functions return `i32` instead of `usize`:

https://github.com/michal-z/zig-gamedev/blob/e31e3c280e0e00d3a15f56a4364a506169a5ebe5/libs/zgui/src/gui.zig#L3465-L3480

I'm down to implement this if it seems reasonable, but I wanted to check in with you first. Should we keep e.g. `getVertexBufferLength` and `getVertexBufferData` and add a `getVertexBufferSlice`, or remove the data/len ones entirely? Does a change like this fit with the general philosophy of the zgui bindings?

I also wonder if they need to be `const` pointers - it seems that it's possible (and useful) to modify this data, for example translate all vertices so they follow a camera.

chip2n:
Making DrawList buffer data non-const and adding helpers for obtaining slices into the buffer data.

Fixes #372

### DIFF
--- a/libs/zgui/src/gui.zig
+++ b/libs/zgui/src/gui.zig
@@ -3468,19 +3468,19 @@ pub const DrawList = *opaque {
     pub const getVertexBufferLength = zguiDrawList_GetVertexBufferLength;
     extern fn zguiDrawList_GetVertexBufferLength(draw_list: DrawList) i32;
     pub const getVertexBufferData = zguiDrawList_GetVertexBufferData;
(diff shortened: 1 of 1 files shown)


hazeycode:
Thanks!

PR #470, opened 2024-02-02
WIP - fixes for zig master

kamidev:
Since we are currently using a designated zig version, this is WIP. But it makes tests pass and sample apps run using zig 0.12.0-dev.2613+0266017b5 (the latest as of today).

There are 3 fixes:

- zgui: fix breaking change in comboFromEnum() function
- zmath: fix breaking changes to std.math.atan2
- zphysics. Comment out failing debugrender tests. Do we need them? Could they be simpler?


michal-z:
Just a note regarding `comboFromEnum()` function. This function was contributed by someone and it is actually used in his project. Also, it is pretty useful. I think it would be good to try to fix it.

kamidev:
> Just a note regarding `comboFromEnum()` function. This function was contributed by someone and it is actually used in his project. Also, it is pretty useful. I think it would be good to try to fix it.

Thanks for the info! I'll try to fix it then.

Srekel:
A few things I had to fix for latest Zig:

`std.os.exit(1); -> std.process.exit(1);`

`pub const BodyDrawFilterFunc = *const fn (*const Body) align(BodyDrawFilterFuncAlignment) callconv(.C) bool;`
remove alignment

`ContainerLayout.Auto -> ContainerLayout.auto`

kamidev:
This PR can be deleted. There is a more recent attempt here: https://github.com/zig-gamedev/zig-gamedev/pull/570

PR #472, opened 2024-02-02
[zglfw][BREAKING] Window API fixes/tweaks

hazeycode:


hazeycode:
Not worth breaking API for

PR #474, opened 2024-02-03
ZPM compatibilty 

hazeycode:
Package products (those consumed via a `std.Build.Dependency`) must only be declared once.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -270,7 +270,7 @@ fn testsWindows(
     optimize: std.builtin.OptimizeMode,
     test_step: *std.Build.Step,
 ) void {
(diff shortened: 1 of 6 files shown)


hazeycode:
This is not strictly required to support ZPM. Only if package consumers want to use `std.Build.Dependency` package interface only. But this (see https://github.com/zig-gamedev/zig-gamedev/pull/474/commits/f17c066830e88942de0cb8fa9b99b0f7d80ed6df) requires that we tie tests to particular instantiations of a libs `Package`, which is not ideal if we want to run tests with different package build options.
I'll think about this more, closing for now.

PR #477, opened 2024-02-04
zopengl: fixed the order of the fields in the wrapper clear Bitfield

loic-o:
It appears to me that the fields in the clear function Bitfield are slightly out of order so this PR attempts to fix that.
If I have missed something, or more is needed for this PR, please let know as I will be happy to correct anything.

As reference:
pub const DEPTH_BUFFER_BIT = 0x00000100;
pub const STENCIL_BUFFER_BIT = 0x00000400;
pub const COLOR_BUFFER_BIT = 0x00004000;

Thanks!

### DIFF
--- a/libs/zopengl/src/wrapper.zig
+++ b/libs/zopengl/src/wrapper.zig
@@ -1520,11 +1520,11 @@ pub fn Wrap(comptime bindings: anytype) type {
         // pub var clear: *const fn (mask: Bitfield) callconv(.C) void = undefined;
         pub fn clear(mask: packed struct(Bitfield) {
             __unused1: u8 = 0,
-            color: bool = false,
-            __unused2: u1 = 0,
             depth: bool = false,
-            __unused3: u3 = 0,
+            __unused2: u1 = 0,
             stencil: bool = false,
+            __unused3: u3 = 0,
+            color: bool = false,
             __unused4: u17 = 0,
         }) void {
             bindings.clear(@bitCast(mask));


hazeycode (changes requested):
Thanks, nice spot!

I think we should also add a test for this to verify it's correct:

```zig
        pub const ClearFlags = packed struct(Bitfield) {
            __unused1: u8 = 0,
            depth: bool = false,
            __unused2: u1 = 0,
            stencil: bool = false,
            __unused3: u3 = 0,
            color: bool = false,
            __unused4: u17 = 0,

            test {
                try std.testing.expectEqual(
                    @clz(@bitReverse(@as(Bitfield, DEPTH_BUFFER_BIT))),
                    @bitOffsetOf(ClearFlags, "depth"),
                );
                try std.testing.expectEqual(
                    @clz(@bitReverse(@as(Bitfield, STENCIL_BUFFER_BIT))),
                    @bitOffsetOf(ClearFlags, "stencil"),
                );
                try std.testing.expectEqual(
                    @clz(@bitReverse(@as(Bitfield, COLOR_BUFFER_BIT))),
                    @bitOffsetOf(ClearFlags, "color"),
                );
            }
        };

        // pub var clear: *const fn (mask: Bitfield) callconv(.C) void = undefined;
        pub fn clear(mask: ClearFlags) void {
            bindings.clear(@bitCast(mask));
        }
```

Also, we need to use `refAllDeclsRecursive` instead of `refAllDecls` at the bottom of zopengl.zig for this test to get run

hazeycode (approved):

PR #478, opened 2024-02-05
DOC/README.md: improved explanation how to get started

guidoschmidt:
Hey I recently update zig-gamedev to the latest version in my own project and found myself struggeling to get it working again. Therefore I added a small section ot the README.md to explain how to get started, might be helpful for others to get started. Happy to get any feedback if that's the *right* way of doing it 🙃

thanks for all the great work on this project btw. 🙏🏽

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,83 @@ We recommend [zigup](https://github.com/marler8997/zigup) for managing compiler 
 | macOS x86_64    | [zig-macos-x86_64-0.12.0-dev.2063+804cee3b9.tar.xz](https://ziglang.org/builds/zig-macos-x86_64-0.12.0-dev.2063+804cee3b9.tar.xz) |
 | macOS aarch64   | [zig-macos-aarch64-0.12.0-dev.2063+804cee3b9.tar.xz](https://ziglang.org/builds/zig-macos-aarch64-0.12.0-dev.2063+804cee3b9.tar.xz) |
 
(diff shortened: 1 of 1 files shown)


hazeycode:
Hi! I was hesitant to add something like this because there are different ways to consume the project.

As it stands, these instructions are too specific. Git is not required if the repo is downloaded as an archive.

But it would be good to have brief instructions for adding the libs to your build.zig.zon (since this is now required) and show how to consume them.

Some considerations for the future (doesn't matter right now):
1. Our `package` interface is somewhat optional. It is (or will be with some minor modifications) possible to consume all package products via `std.Build.Dependency` instead and we may retire our `package` interface if it causes problems or becomes completely redundant.
2. Vendoring libs is optional. Local paths are currently used exclusively just because we arn't yet publishing packages to be fetched via URL. User may choose to vendor dependencies or fetch from remote.

guidoschmidt:
@hazeycode nice thanks for the detaild feedback. Indeed, it's very specific to how I'm using the libraries atm. I might generalize it more. Also I still haven't worked much with zigs zon format and package manager.

hazeycode:
I pushed https://github.com/zig-gamedev/zig-gamedev/commit/6f890ad65fd619aa91e739d4b28b93c726e171b5, feel free to reopen this if I've missed something.

guidoschmidt:
Awesome @hazeycode, I think this little section will already help other devs understand how to use zig-gamedev 👍🏽 thtanks for the push

PR #480, opened 2024-02-07
Emscripten support

hazeycode:
WIP Emscripten support


hazeycode:
Closing this because it's old and because zemscripten lib has been added in [this commit](https://github.com/zig-gamedev/zig-gamedev/commit/a7c35b92f032a22bbbeac83d74821ee251862ba9). Will follow up with PRs for emscripten support in libs and samples.

PR #482, opened 2024-02-11
zgui: Add Drag And Drop API

devBoi76:
This PR adds the ImGui Drag And Drop API, which was missing, to zgui.

I tested some basic functionality (essentially reimplemented the example in `imgui_demo.cpp` [here](https://github.com/ocornut/imgui/blob/master/imgui_demo.cpp#L2353)) and it seems to work.

I'm not sure whether the C++ code is formatted correctly, I did it by hand but I don't know if I'm supposed to run it through some formatter.

### DIFF
--- a/libs/zgui/src/gui.zig
+++ b/libs/zgui/src/gui.zig
@@ -3399,7 +3399,7 @@ const Payload = extern struct {
     source_id: c_uint = 0,
     source_parent_id: c_uint = 0,
     data_frame_count: c_int = -1,
-    data_type: [32 + 1]c_char,
+    data_type: [32:0]c_char,
     preview: bool = false,
     delivery: bool = false,
 
@@ -3430,7 +3430,7 @@ pub fn beginDragDropSource(flags: DragDropFlags) bool {
     return zguiBeginDragDropSource(flags);
 }
 pub fn setDragDropPayload(payload_type: [*:0]const u8, data: []const u8, cond: Condition) bool {
-    return zguiSetDragDropPayload(payload_type, data.ptr, data.len, cond);
+    return zguiSetDragDropPayload(payload_type, @alignCast(@ptrCast(data.ptr)), data.len, cond);
 }
 pub fn endDragDropSource() void {
     zguiEndDragDropSource();
@@ -3448,7 +3448,7 @@ pub fn getDragDropPayload() ?*Payload {
     return zguiGetDragDropPayload();
 }
 extern fn zguiBeginDragDropSource(flags: DragDropFlags) bool;
-extern fn zguiSetDragDropPayload(type: [*:0]const u8, data: *anyopaque, sz: usize, cond: Condition) bool;
+extern fn zguiSetDragDropPayload(type: [*:0]const u8, data: *const anyopaque, sz: usize, cond: Condition) bool;
 extern fn zguiEndDragDropSource() void;
 extern fn zguiBeginDragDropTarget() bool;
 extern fn zguiAcceptDragDropPayload(type: [*:0]const u8, flags: DragDropFlags) [*c]Payload;
--- a/libs/zgui/src/gui.zig
+++ b/libs/zgui/src/gui.zig
@@ -3441,7 +3441,7 @@ pub fn acceptDragDropPayload(payload_type: [*:0]const u8, flags: DragDropFlags) 
 pub fn endDragDropTarget() void {
     zguiEndDragDropTarget();
 }
-pub fn getDragDropPayload() *Payload {
+pub fn getDragDropPayload() ?*Payload {
     return zguiGetDragDropPayload();
 }
 extern fn zguiBeginDragDropSource(flags: DragDropFlags) bool;
--- a/libs/zgui/src/gui.zig
+++ b/libs/zgui/src/gui.zig
@@ -3371,6 +3371,88 @@ extern fn zguiGetMouseDragDelta(button: MouseButton, lock_threshold: f32, delta:
 extern fn zguiResetMouseDragDelta(button: MouseButton) void;
 //--------------------------------------------------------------------------------------------------
 //
+// Drag and Drop
+//
+//--------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


hazeycode:
Looks good, thanks! I'll review properly later today.

We don't use anything for C++ formatting currently. Maybe we should but that's another topic.

libs/zgui/src/gui.zig
```diff
@@ -3371,6 +3371,88 @@ extern fn zguiGetMouseDragDelta(button: MouseButton, lock_threshold: f32, delta:
...
+
+    _padding1: u18 = 0,
+};
+
+const Payload = extern struct {
+    data: *anyopaque = null,
+    data_size: i32 = 0,
+    source_id: u32 = 0,
+    source_parent_id: u32 = 0,
+    data_frame_count: i32 = -1,
```
hazeycode, line 3400: The `i32`s should be `c_int`s and the `u32`s should be `c_uint`, for maximum portability.

I do see that we're already using explicitly 32 bit ints elsewhere in these bindings. I think this is incorrect and they should be changed to match the underlying API (outside of this PR)

libs/zgui/src/gui.zig
```diff
@@ -3371,6 +3371,88 @@ extern fn zguiGetMouseDragDelta(button: MouseButton, lock_threshold: f32, delta:
...
+    /// `pub fn isDelivery(payload: *const Payload) bool;
+    pub const isDelivery = zguiImGuiPayload_IsDelivery;
+    extern fn zguiImGuiPayload_IsDelivery(payload: *const Payload) bool;
+};
+
+pub fn beginDragDropSource(flags: DragDropFlags) bool {
+    return zguiBeginDragDropSource(flags);
+}
+pub fn setDragDropPayload(payload_type: [*:0]const u8, data: *anyopaque, sz: usize, cond: Condition) bool {
+    return zguiSetDragDropPayload(payload_type, data, sz, cond);
```
hazeycode, line 3430: We can capitalise on Zig's slice type here, i.e.
```zig
pub fn setDragDropPayload(payload_type: [*:0]const u8, data: []const u8, cond: Condition) bool {
     return zguiSetDragDropPayload(payload_type, data.ptr, data.len, cond);
```

To call this with something other than bytes would look like:
```zig
_ = zgui.setDragDropPayload("MyPayload", std.mem.sliceAsBytes(my_payload));
```

I think this would be nice because it prevents passing the wrong size. What do you think?

libs/zgui/src/gui.zig
```diff
@@ -3371,6 +3371,88 @@ extern fn zguiGetMouseDragDelta(button: MouseButton, lock_threshold: f32, delta:
...
+    data_size: i32 = 0,
+    source_id: u32 = 0,
+    source_parent_id: u32 = 0,
+    data_frame_count: i32 = -1,
+    data_type: [32 + 1]u8,
+    preview: bool = false,
+    delivery: bool = false,
+
+    /// `pub fn init() Payload`
+    pub const init = zguiImGuiPayload_Init;
```
hazeycode, line 3406: We should careful here. User's might expect `zgui.Payload.init()` to behave the same way as the C++ constructor.

I think we should call `Clear` here to mirror the C++ API i.e.
```zig
pub fn init() Payload {
    var payload = Payload{};
    payload.clear();
    return payload;
}
```

hazeycode (changes requested):

libs/zgui/src/gui.zig
```diff
@@ -3371,6 +3371,88 @@ extern fn zguiGetMouseDragDelta(button: MouseButton, lock_threshold: f32, delta:
...
+    source_allow_null_id: bool = false,
+    source_extern: bool = false,
+    source_auto_expire_payload: bool = false,
+
+    _padding0: u4 = 0,
+
+    accept_before_delivery: bool = false,
+    accept_no_draw_default_rect: bool = false,
+    accept_no_preview_tooltip: bool = false,
+    accept_peek_only: bool = false,
```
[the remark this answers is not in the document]
devBoi76: You're right, I must have missed it. Done.

libs/zgui/src/gui.zig
devBoi76: Alright, I tried to match what's already here and a bunch of stuff was using zig types. Good note on the portability.

libs/zgui/src/gui.zig
hazeycode: My mistake, that should of been
```zig
    return zguiSetDragDropPayload(payload_type, @ptrCast(data.ptr), data.len, cond);
```

hazeycode:
By the way, I have created an issue to address ABI compat on non 32-bit word systems: https://github.com/zig-gamedev/zig-gamedev/issues/483

libs/zgui/src/gui.zig
```diff
@@ -3399,7 +3399,7 @@ const Payload = extern struct {
     source_id: c_uint = 0,
     source_parent_id: c_uint = 0,
     data_frame_count: c_int = -1,
-    data_type: [32 + 1]c_char,
+    data_type: [32:0]c_char,
```
hazeycode, line 3402: nice!

devBoi76:
Actually now that I'm looking at the `Payload` struct I see that `data_type` is a `[32:0]c_char`. Should the functions that take the type as its parameter somehow validate the size zig-side or do we leave that up to imgui?

devBoi76:
I think I'll add a doc comment that clarifies this

hazeycode:
> Actually now that I'm looking at the `Payload` struct I see that `data_type` is a `[32:0]c_char`. Should the functions that take the type as its parameter somehow validate the size zig-side or do we leave that up to imgui?

Good question. I think we should keep the bindings shallow and let Dear Imgui do what it does, which in this case appears to be just a `strcmp` which will determine that 2 null terminated strings of different lengths are never equal.

hazeycode (approved):
Thanks! Are you happy for me to merge this now?

devBoi76:
Yup! 👍

Issue #487, opened 2024-02-14
zglfw is missing binding with glfwSetCharCallback


robinheydon:
There are bindings for other callback functions such as `glfwSetKeyCallback`, but the `glfwSetCharCallback` function is missing.

PR #488, opened 2024-02-14
Added binding to glfwSetCharCallback

robinheydon:
Very single PR - just added binding to the `glfwSetCharCallback` function from glfw. This function sends codepoints to the callback function when a key is pressed. e.g. when you press the compose key and they `<` and `<` again, you'd get the codepoint 0x00AB sent to the callback `«`.

This should resolve [Issue 487](https://github.com/zig-gamedev/zig-gamedev/issues/487)


hazeycode:
Thanks!

Issue #491, opened 2024-02-14
zgui does not expose the ability to check that the text has been captured


robinheydon:
In the imgui (docs)[https://github.com/ocornut/imgui/blob/master/docs/FAQ.md#q-how-can-i-tell-whether-to-dispatch-mousekeyboard-to-dear-imgui-or-my-application] the way to check that imgui has done something useful with the mouse and keyboard is to check three functions: `io.WantCaptureMouse`, `io.WantCaptureKeyboard` and `io.WantTextInput`.

The first two are exposed in zgui using the `zgui.io.getWantCapture*` functions. The last does not appear to be exposed.

PR #492, opened 2024-02-14
[zgui] Added getWantTextInput

robinheydon:
Added `zguiIoGetWantTextInput` into zgui.cpp, and added binding in gui.zig with `getWantTextInput`.

This should resolve [Issue 491](https://github.com/zig-gamedev/zig-gamedev/issues/491)


hazeycode:
Thanks!

PR #496, opened 2024-02-15
Add padding for ImGuiWindowFlags 17 value

michaelbartnett:
There's a hole in the enumerated values for ImGuiWindowFlags, it skips from `1<<16` to `1<<18`, so anything after always_use_window_padding wasn't working properly. Noticed because `.{ .unsaved_document = true }` wasn't making the dot appear in the title bar.

https://github.com/zig-gamedev/zig-gamedev/blob/73ccc8e894c22319281f6de150e125d2493e703a/libs/zgui/libs/imgui/imgui.h#L975-L976


hazeycode:
Thanks!

PR #501, opened 2024-02-17
extract zcgltf from zmesh into it's own lib

hazeycode:
TODO
- [ ] Zig allocator interface (overrides MemoryOptions)


hazeycode:
Plan is to replace zcgltf with a pure zig impl. Also see c9afa6c4d2b95b808e8c6f21124d61abf61ca5f4

PR #504, opened 2024-02-22
Added TracyAllocator to ztracy

robinheydon:
Added TracyAllocator to ztracy.

In the not-used case, the allocator just returns the child_allocator, a true zero-cost abstraction.

In the used case, it returns a new allocator that wraps the child_allocators alloc, resize, and free functions, recording the Alloc, Free/Alloc, or Free with tracy if they succeed or recording a message if not.


hazeycode:
Thanks!

PR #505, opened 2024-02-24
zglfw: upgrade glfw srcs to ver 3.4

hazeycode:
- [x] test macOS
- [x] test Windows
- [x] test Linux


hazeycode:
These changes are currently causes frame_pacing_wgpu to crash on X11 when switching to fullscreen

hazeycode:
@Pyrolistical could you check that frame_pacing_wgpu still works with this on Windows? For some reason I can't get that particular sample to build on Windows (seems like a Zig bug)

hazeycode:
frame_pacing_wgpu crash fixed by explicitly setting client_api window hint to no_api https://github.com/zig-gamedev/zig-gamedev/pull/505/commits/90d30efdbd67d514e45a234e3ceadc85649b3540

hazeycode:
> @Pyrolistical could you check that frame_pacing_wgpu still works with this on Windows? For some reason I can't get that particular sample to build on Windows (seems like a Zig bug)

nvm, I got it building and running and it works as before

PR #506, opened 2024-02-25
Remove redundant comptime in zcgltf test

btipling:


btipling:
oh much better name for the PR

PR #509, opened 2024-02-25
std.math.atan2 no longer takes a type arg

btipling:
https://ziglang.org/documentation/master/std/#A;std:math.atan2
https://ziglang.org/documentation/master/std/src/std/math/atan2.zig.html#L33

This fixes `zig build instanced_pills_wgpu-run`


hazeycode:
The main branch tracks a pinned version (currently 0.12.0-dev.2063+804cee3b9) so I've retargeted this to our unstable branch which is intended for changes that are incompatible with / ahead of the pinned version.

Which version version are you using?

PR #512, opened 2024-02-29
zflecs: Add a few more functions to the bindings

Hejsil:
Need `ecs_set_threads`. Tought I might as well add everything after `ecs_progress` and down from the header


hazeycode:
Thanks!

Issue #467, opened 2024-01-29
Tests not linking SDL on Windows


hazeycode:
Fails here -> https://github.com/zig-gamedev/zig-gamedev/actions/runs/7700805077 with commit https://github.com/zig-gamedev/zig-gamedev/commit/028522e0ebdab0b2290d3dcce5f34a1805808ba6

Undo this commit -> https://github.com/zig-gamedev/zig-gamedev/commit/f0a2b026d86401494fcaa2d74ed73b370491810f

Pyrolistical:
https://github.com/zig-gamedev/zig-gamedev/issues/510 is blocked by this as openvr comes with a `.dll` that needs to linked in the same way as sdl

Pyrolistical:
a workaround is to copy the .dll to where the .exe is

Pyrolistical:
the symptom is exit code 53 if the .dll is not found by the linked .exe

Pyrolistical:
Is this related to https://github.com/ziglang/zig/issues/5289 ?

hazeycode:
> Is this related to [ziglang/zig#5289](https://github.com/ziglang/zig/issues/5289) ?

I'm not sure. But it appears to have been fixed by https://github.com/zig-gamedev/zig-gamedev/pull/518.

However there is a regression on Windows. I tried reapplying your `setCwd` change (https://github.com/Pyrolistical/zig-gamedev/commit/4aa6a7acf5d3bd093ed9d25c1aa929068ece0d89) but it doesn't work now. Sorry, I will try to fix it.

PR #513, opened 2024-03-01
[zsdl] handle windows case for #467 by setting cwd instead of rpath

Pyrolistical:


libs/zsdl/src/sdl2.zig
```diff
@@ -125,6 +125,10 @@ pub fn getVersion() Version {
 }
 extern fn SDL_GetVersion(version: *Version) void;
 
+test "compiled version should be same as linked version" {
+    try std.testing.expectEqual(VERSION, getVersion());
+}
```
hazeycode, line 130: This is useful, but I think it might be too restrictive and/or confusing. If a user decides to discard the bundled SDL lib, the tests should still work.

I have been considering making the bundled SDL lib optional. SDL is fairly ubiquitous and ships backwards compatible improvements. IMO, the only real advantage to bundling it is for cross-compilation and this use case would be better served with stubs.

Pyrolistical: I only added it so I could check tests actually work as there were none. We could just remove this or soften it to only check major

hazeycode: Makes sense. How about moving it into it's own test step that can be run separately?

Pyrolistical: Ya I can pull it out

hazeycode:
Thanks!

PR #514, opened 2024-03-03
wip: fixup zmesh build script to support being used a dependency

kcbanner:
This is just a draft to demonstrate the pattern of changes that I would be making, before proceeding with the rest.

Currently, there is unfortunately a limitation of the build system that requires any artifact that will be accessed via the `Dependency` to be installed. For example, in the zmesh case, this means zmesh.lib and zmesh-tests.exe end up in the zig-out folder. 

There is some discussion about that problem here: https://github.com/ziglang/zig/pull/19017#issuecomment-1975193399 and here: https://discord.com/channels/605571803288698900/1145919498801655839/1213910624841506857.

I may hold on off proceeding with these changes until that's cleared up.


kcbanner:
Closing this in favour of https://github.com/zig-gamedev/zig-gamedev/pull/515 for now.

Issue #516, opened 2024-03-04
macos 11.7 amd64(x86_64) build error


Angluca:
```zsh
zig version
0.12.0-dev.3142+9d500bda2

zig build
warning(zgpu):
---------------------------------------------------------------------------

Dawn/WebGPU binary for this target is not available.

Following targets are supported:

x86_64-windows-gnu
x86_64-linux-gnu
x86_64-macos.12.0.0-none
aarch64-linux-gnu
aarch64-macos.12.0.0-none

---------------------------------------------------------------------------

install
└─ minimal_zgui_glfw_gl
   └─ install minimal_zgui_glfw_gl
      └─ zig build-exe minimal_zgui_glfw_gl Debug native
         └─ zig build-lib zgui Debug native 1 errors
error: unable to check cache: stat file '/Users/xxx/Zigs/zig-gamedev/Users/xxx/Zigs/zig-gamedev/libs/zgui/libs/imgui/imgui.cpp' failed: FileNotFound
error: the following command failed with 1 compilation errors:
[...]

Issue #511, opened 2024-02-28
Building and running samples error (probably path related)


noncom:
I just cloned the repo, and I can't run the first recommended example.
When trying to run it I'm getting this:

```powershell
Microsoft Windows [Version 10.0.19045.4046]
(c) Microsoft Corporation. All rights reserved.

D:\work\zig\zig-gamedev>zig build physically_based_rendering_wgpu-run
physically_based_rendering_wgpu-run
└─ run physically_based_rendering_wgpu
   └─ physically_based_rendering_wgpu
      └─ install physically_based_rendering_wgpu
         └─ zig build-exe physically_based_rendering_wgpu Debug native
            └─ zig build-lib zmesh Debug native 1 errors
error: unable to check cache: stat file 'D:\work\zig\zig-gamedev\D:\work\zig\zig-gamedev\libs\zmesh\libs\meshoptimizer\clusterizer.cpp' failed: Unexpected     
error: the following command failed with 1 compilation errors:
[...]

PR #518, opened 2024-03-09
Retire build Package API

hazeycode:
Our build package API served its purpose prior to the introduction of Zig's official package manager. Maintaining a custom package build API while being compatible with ZPM has so far been tricky and with our next pinned version of 0.12.0-dev.3180+83e578a18 becomes ever more problematic. It's time to retire it.

* Removes much inversion-of-control in build system which causes compatibility problems with ZPM
* Unlocks compatibility with next nominated Zig version (12.0-dev.3180+83e578a18) and fixes for #511 and #516
* Leans on Zig Package Manager for caching lib build products
* More flexibility when consuming libs; user can take all or some parts.
* Requires all consumers declare vendor'd library paths in their build.zig.zon

### Issues
- `minimal_sdl_gl-run` crashes on Linux
- zsdl tests not linking on Windows (disabled)
- Artifacts must be installed using `b.installArtifact` in order to be exposed to `Dependency` consumers.
- `error: Overflow` building zgpu tests and samples with 12.0-dev.3180+83e578a18

### Example

Instead of this:
```zig
const zgui = @import("../path/to/zgui/build.zig").package(b, .{ .backend = .glfw_wgpu });
zgui.link(exe);
```
now we do this
```zig
const zgui = b.dependency("zgui", .{ .backend = .glfw_wgpu });
exe.root_module.addImport("zgui", zgui.module("root"));
exe.linkLibrary(zgui.artifact("imgui"));
```

CC @kcbanner @Pyrolistical @Srekel @michal-z


Srekel:
Sounds good!

kcbanner:
Nice, this looks great! Thanks for doing this.

hazeycode:
I've merged this with some outstanding issues (as detailed in description). I think they don't need to hold back merging this and can be resolved as part of following work to move to Zig 0.12.0-dev.3180+83e578a18

PR #519, opened 2024-03-10
added zgui glfw+dx12 backend and sample

Pyrolistical:
Looking for feedback on how to improve this call:
https://github.com/zig-gamedev/zig-gamedev/pull/519/files#diff-5d56ea36cb7e424a98b9756f43f8ed65fac2bbe159384ccecc0df12dc3c5d985R63

I didn't want to add [`zwin32.d3d12.CPU_DESCRIPTOR_HANDLE`](https://github.com/zig-gamedev/zig-gamedev/blob/193a1d278f2b4970745c171056c6480bca64bbf5/libs/zwin32/src/d3d12.zig#L64-L66) as a dep in zgui, so I ended up using a `@bitCast`.

Ditto for [`zwin32.dxgi.FORMAT`](https://github.com/zig-gamedev/zig-gamedev/blob/193a1d278f2b4970745c171056c6480bca64bbf5/libs/zwin32/src/dxgi.zig#L23).

The alternative is just to add some build conditions to only pull in zwin32 when using glfw_dx12 backend

### DIFF
--- a/libs/zgui/build.zig
+++ b/libs/zgui/build.zig
@@ -4,6 +4,7 @@ pub const Backend = enum {
     no_backend,
     glfw_wgpu,
     glfw_opengl3,
+    glfw_dx12,
     win32_dx12,
 };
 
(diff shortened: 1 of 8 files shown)


libs/zgui/build.zig
```diff
@@ -148,6 +149,18 @@ pub fn package(
...
         },
+        .glfw_dx12 => {
+            const zglfw = b.dependency("zglfw", .{});
+            zgui_c_cpp.addIncludePath(.{ .path = zglfw.path("libs/glfw/include").getPath(b) });
+            zgui_c_cpp.addCSourceFiles(.{
+                .files = &.{
+                    thisDir() ++ "/libs/imgui/backends/imgui_impl_glfw.cpp",
+                    thisDir() ++ "/libs/imgui/backends/imgui_impl_dx12.cpp",
+                },
+                .flags = &(cflags.* ++ .{"-DIMGUI_IMPL_OPENGL_LOADER_CUSTOM"}),
```
hazeycode, line 160: I don't think we need this flag

Pyrolistical: oops. fixed and rebased

PR #520, opened 2024-03-12
zflecs: Fix PerTypeGlobalVar for latest zig

Hejsil:
As of <https://github.com/ziglang/zig/pull/19190> `PerTypeGlobalVar` no longer generates a unique global variable for each `in_type`. The fix is to capture `in_type` in the returned struct.

This should be backwards compatible with the version of zig `zig-gamedev` targets.


hazeycode:
Thanks!

PR #525, opened 2024-03-21
zgui: Initial support for ImGUI Test Engine

OndraVoves:
This is initial support for ImGUI test engine. It is not ideal but works ;)
Btw. with `zglfw.windowHintTyped(.visible, false);` you can run `zgui` and `zgpu` in GitHub Actions, but only on macOS and windows, linux need OSMesa enabled in glfw.


hazeycode (approved):
Nice!

PR #532, opened 2024-03-24
Native file dialog extended - support

OndraVoves:
Another day another PR 😄 

This solution works on linux (test on SteamDeck) with bundled dbus in this PR but if I try used system-sdk on macOS its failed. For now is working  on all native platforms (Windows/linux/macos) but problem will be probably with cross-compile because that task need system-sdk libs.


hazeycode:
FWIW I have some NFD-extended bindings here that I'm using in a project https://github.com/hazeycode/znfd/blob/main/src/nfd.zig
It handles UTF-16 mode and doesn't involve translate-c

OndraVoves:
Nice 👍  we can combine your lib + dbus in system-sdk and zig-gamedev has new lib ;)

OndraVoves:
Closing this PR but in near feature a make new try with @hazeycode nfd + dbus in systemsdk

PR #534, opened 2024-03-26
fixed comboFromEnum assuming enum int value is consecutive

Pyrolistical:
This PR enhances `zgui.comboFromEnum` to work with sparse enums such as:
```zig
const SparseEnum = enum(i32) {
    first = 10,
    second = 100,
    third = 1000,
};
```


hazeycode:
Nice!

Issue #535, opened 2024-03-28
Tracy Zone Name Mismatch


jcalabro:
Creating an issue for discussion separate from the PR (#530)

When using latest zig (or the most recent Mach nominated Zig), all of the zones we emit to Tracy take one name (it appears to be the first zone I use):

![image](https://github.com/zig-gamedev/zig-gamedev/assets/8205547/ad93433f-f8f8-4336-9f1c-4817419900bd)

I did some poking around and noticed that if I heap allocate the [___tracy_source_location_data](https://github.com/zig-gamedev/zig-gamedev/blob/078c568d2b37ea613c194641797a1be21ad6fc39/libs/ztracy/src/ztracy.zig#L322-L331) object, it works fine (though heap allocating here is not super awesome). I also noticed that if I just don't use the local `static` struct in [this commit](https://github.com/jcalabro/zig-gamedev/commit/7c23e43ff0bf0b63f58c5b46e257f27a6afe7d52) for example, it also works just fine.

![image](https://github.com/zig-gamedev/zig-gamedev/assets/8205547/ae0f75b3-4895-4126-96b5-6625ec88831c)

I am going to be totally honest, I'm not sure why this is the case. There's also a lot of Zig commits to dig through to try and find out what changed. Any chance you know off the top of your head? Otherwise, I can float it in Discord.

hazeycode:
@jcalabro looks like @robinheydon identified the problem here  https://github.com/zig-gamedev/zig-gamedev/pull/536/commits/6b44e904fabc2269a7dbfcb4ecc842c1176e085f but I haven't verified yet

jcalabro:
Just tried the fix and it works great for me. Thanks all! Will do some reading on those issues in the comments :eyes:

hazeycode:
Nice! Thanks for checking. But let's leave this open until this change is merged.

PR #536, opened 2024-03-28
Flecs identifiers

robinheydon:
This adds a number of flecs identifiers into zig-gamedev. This is done either by just adding the external symbol definition or by using `ecs_id` (which uses `@extern` to reference the symbol created by the C macro used to create the symbol).

This also changes the interface for flecs.init to take an allocator, and then use that allocator within flecs. This allows the use of tracy allocator or a logging allocator to watch what flecs is doing with memory.

Added `componentWithOptions` for defining a component with additional options, such as an explicit name / symbol rather than just relying on `@typeName`. Also added `tagWithOptions` for defining a tag with additional options.

Also added `setMonitor` into the zglfw Window struct.

Also fixed ztracy allocator to the latest zig allocator interface.


hazeycode:
@Srekel Could you help review this?

hazeycode:
@robinheydon which verison of Zig are you using/targeting?

hazeycode (changes requested):
zflecs/src/tests.zig needs updating

libs/zflecs/build.zig
```diff
@@ -13,6 +13,7 @@ pub fn build(b: *std.Build) void {
         .target = target,
         .optimize = optimize,
     });
+    b.installArtifact(flecs);
```
hazeycode, line 16: I'm not  sure what's going on with the diff, but `installArtifact(flecs)` is already being called on main branch, albeit not on this exact line. Can you try rebasing this branch?

libs/ztracy/build.zig
```diff
@@ -24,12 +24,13 @@ pub fn build(b: *std.Build) void {
...
     const options_module = options_step.createModule();
 
-    _ = b.addModule("root", .{
+    const ztracy = b.addModule("root", .{
         .root_source_file = .{ .path = "src/ztracy.zig" },
         .imports = &.{
             .{ .name = "ztracy_options", .module = options_module },
         },
     });
+    ztracy.addIncludePath(.{ .path = "libs/tracy/tracy" });
```
hazeycode, line 33: Likewise here (changes present on main not showing in diff)

libs/ztracy/src/ztracy.zig
```diff
@@ -321,6 +321,13 @@ const tracy_full = struct {
         // struct.
         const static = struct {
             var loc: c.___tracy_source_location_data = undefined;
+
+            // Ensure that a unique struct type is generated for each unique `src`. See
+            // https://github.com/ziglang/zig/issues/18816
+            comptime {
+                // https://github.com/ziglang/zig/issues/19274
+                _ = @sizeOf (@TypeOf (src));
```
hazeycode, line 329: Is this the solution to https://github.com/zig-gamedev/zig-gamedev/issues/535?

hazeycode (changes requested):
Please use `zig fmt .` in repo root

Srekel:
> @Srekel Could you help review this?

Yeah I'll take a look after your initial concerns have been addressed :)

libs/zflecs/build.zig
robinheydon: yes, will do.

libs/ztracy/build.zig
robinheydon: strange... I'll have a look

Srekel:
FWIW the idea behind this seems good and probably worth revisiting for flecs 4, but I'm not sure why it's touching zglfw and tracy. If those changes are needed for those libs, would be better to do as a separate PR.

hazeycode:
Closing because this is quite out of date. Please reopen on zflecs repo.

PR #407, opened 2023-09-16
Update imgui to 1.89.9-docking and implot to 0.16

dstrachan:

PR #537, opened 2024-03-28
[zgui][BREAKING] Update ImGui + test engine to 1.90.4-docking and ImPlot 0.17

OndraVoves:
Thanks @dstrachan for initial work in #407 👍 
Btw problem with tiny scrollbar is not probably bug. In example size is set to 6 but in imgui default is 14, but maybe i miss something.

### beginChild and beginChildId
Change
```zig
zgui.beginChild("FOO", .{ .border = true }))
```
To
```zig
zgui.beginChild("FOO",.{ .child_flags = .{ .border = true } })
```

### DIFF
--- a/samples/gui_test_wgpu/src/gui_test_wgpu.zig
+++ b/samples/gui_test_wgpu/src/gui_test_wgpu.zig
@@ -106,7 +106,7 @@ fn create(allocator: std.mem.Allocator, window: *zglfw.Window) !*DemoState {
 
     style.window_min_size = .{ 320.0, 240.0 };
     style.window_border_size = 8.0;
-    style.scrollbar_size = 6.0;
+    style.scrollbar_size = 12.0;
     {
         var color = style.getColor(.scrollbar_grab);
         color[1] = 0.8;
--- a/samples/gui_test_wgpu/src/gui_test_wgpu.zig
+++ b/samples/gui_test_wgpu/src/gui_test_wgpu.zig
@@ -106,7 +106,7 @@ fn create(allocator: std.mem.Allocator, window: *zglfw.Window) !*DemoState {
 
     style.window_min_size = .{ 320.0, 240.0 };
     style.window_border_size = 8.0;
-    style.scrollbar_size = 12.0;
+    style.scrollbar_size = 6.0;
     {
         var color = style.getColor(.scrollbar_grab);
         color[1] = 0.8;
--- a/libs/zgui/README.md
+++ b/libs/zgui/README.md
@@ -1,4 +1,4 @@
-# zgui v0.1.0 - dear imgui bindings
+# zgui v0.2.0 - dear imgui bindings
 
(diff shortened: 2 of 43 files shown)


hazeycode:
Nice work! I'll review this properly ASAP.

hazeycode (commented):
I think we want an `enable_docking` option that defines `IMGUI_HAS_DOCK` also? Default to true?

samples/gui_test_wgpu/src/gui_test_wgpu.zig
```diff
@@ -106,7 +106,7 @@ fn create(allocator: std.mem.Allocator, window: *zglfw.Window) !*DemoState {
 
     style.window_min_size = .{ 320.0, 240.0 };
     style.window_border_size = 8.0;
-    style.scrollbar_size = 6.0;
+    style.scrollbar_size = 12.0;
```
hazeycode, line 109: I'm nervous about this because I don't understand why it's changed.

OndraVoves: FIxed. My bad ;)

hazeycode: But without this change, the scroll bar is thinner than before

OndraVoves:
> I think we want an `enable_docking` option that defines `IMGUI_HAS_DOCK` also? Default to true?

On docking branch is defined by imgui itself and it work like normal imgui until you  enable "dock feeature"  by `dock_enable` in `zgui.ConfigFlags`.  (define `IMGUI_HAS_DOCK ` in build.zig is my mistake)

samples/gui_test_wgpu/src/gui_test_wgpu.zig
OndraVoves: Yes because 6 is really thin, default in imgui is `12`. strange i know 😅

hazeycode:
> define `IMGUI_HAS_DOCK ` in build.zig is my mistake

I see `IMGUI_HAS_DOCK` used by preprocessor in imgui.cpp. I think we want to also compile that code?

OndraVoves:
> > define `IMGUI_HAS_DOCK ` in build.zig is my mistake
> 
> I see `IMGUI_HAS_DOCK` used by preprocessor in imgui.cpp. I think we want to also compile that code?

Corect and its defined here https://github.com/zig-gamedev/zig-gamedev/blob/ca91e27873a237a056287974cf487b4ed6c3f85b/libs/zgui/libs/imgui/imgui.h#L30

samples/gui_test_wgpu/src/gui_test_wgpu.zig
hazeycode: It seems about right. Maybe something was fixed upstream. Let's go with it.

OndraVoves:
🎉 NICE 🎉

Pyrolistical:
Hmm, this change broke my local wip with:
```sh
Assertion failed: viewport->RendererUserData == 0 && viewport->PlatformUserData == 0, file ...\libs\zgui\libs\imgui\imgui.cpp, line 15466
```

I am doing something out of the ordinary by re-initing zgui.

PR #538, opened 2024-03-31
fix grammar in readme

terraquad:


copygirl:
The sentence as it is, is grammatically correct.

Issue #352, opened 2023-07-22
[zgui] Weird textInput buffer handling


mifopen:
Hey!
I have noticed some weird behaviour of zgui.textInput that I do not understand.
Let's say we have the following code written in C/C++:
```c
static char text[1024] = "";
static ImGuiInputTextFlags flags = ImGuiInputTextFlags_EnterReturnsTrue;
if(ImGui::InputText("##source", text, IM_ARRAYSIZE(text), flags)) {
  fprintf(stdout, "%s\n", text);
}
```
and now we want to convert it into Zig using zig-gamedev/zgui lib.
```zig
const static = struct {
    var buf: [128]u8 = undefined;
};
if (zgui.inputText(
    "##source",
    .{
        .buf = &static.buf,
        .flags = .{
            .enter_returns_true = true,
        },
    },
)) {
    std.debug.print("{s}\n", .{&static.buf});
}
```

Now if you do this:
1. Enter "hello"
2. Press Enter
3. See "hello" in terminal
4. Focus input again
5. ctrl+a / cmd+a (select text)
6. Type 'x'
7. Press Enter

For the C variant, you will see `x` in the terminal as expected
For the Zig variant, you will see `xllo` in the terminal.

It could be definitely that I don't understand some very basic stuff about system programming/C/Zig as I'm just entering this realm. Could you please help me understand what is going on here?

hazeycode:
Hey @mifopen, it looks like Dear ImGui only write a single terminal 0 after the string it reads, as opposed to zeroing the entire buffer.

Note that in you Zig version, you are trying to print the entire buffer, you could try using [`std.mem.sliceTo`](https://ziglang.org/documentation/master/std/#A;std:mem.sliceTo)

I would recommend also that you zero the entire buffer before calling `inputText`.

We should perhaps consider having `inputText` zero the buffer by default @michal-z ?

mifopen:
Thank you @hazeycode! This makes a lot of sense. I agree that it is not obvious how thin should Zig binding libs be in general. I will close the issue then.

Pyrolistical:
There are 2 different bugs interacting here.

1. `var buf: [128]u8 = undefined;` is being incorrectly zeroed. I think it is caused by https://github.com/ziglang/zig/issues/19538
[...]

PR #541, opened 2024-04-04
[zgui] Fixed buffer type for inputText* to include null sentinel.

Pyrolistical:
The slice length is for maximum text length, whereas the sentinel denotes the current text length.

Added output of current text length, which would have errored before this change since the buffers didn't include a sentinel in their type.

See https://github.com/zig-gamedev/zig-gamedev/issues/352#issuecomment-2038175998

### DIFF
--- a/samples/gui_test_wgpu/src/gui_test_wgpu.zig
+++ b/samples/gui_test_wgpu/src/gui_test_wgpu.zig
@@ -18,6 +18,9 @@ const DemoState = struct {
     font_normal: zgui.Font,
     font_large: zgui.Font,
     draw_list: zgui.DrawList,
+    alloced_input_text_buf: [:0]u8,
+    alloced_input_text_multiline_buf: [:0]u8,
+    alloced_input_text_with_hint_buf: [:0]u8,
 };
 var _te: *zgui.te.TestEngine = undefined;
 
@@ -105,7 +108,6 @@ fn create(allocator: std.mem.Allocator, window: *zglfw.Window) !*DemoState {
     const style = zgui.getStyle();
 
     style.window_min_size = .{ 320.0, 240.0 };
-    style.window_border_size = 8.0;
     style.scrollbar_size = 6.0;
     {
         var color = style.getColor(.scrollbar_grab);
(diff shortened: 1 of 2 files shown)


samples/gui_test_wgpu/src/gui_test_wgpu.zig
```diff
@@ -478,12 +489,30 @@ fn update(demo: *DemoState) !void {
...
             };
-            _ = zgui.inputText("Input text", .{ .buf = static.buf[0..] });
-            _ = zgui.inputTextMultiline("Input text multiline", .{ .buf = static.buf1[0..] });
-            _ = zgui.inputTextWithHint(
-                "Input text with hint",
-                .{ .hint = "Enter your name", .buf = static.buf2[0..] },
-            );
+            zgui.separatorText("static input text");
+            _ = zgui.inputText("Input text", .{ .buf = static.input_text_buf[0..] });
+            _ = zgui.text("length of Input text {}", .{std.mem.len(@as([*:0]u8, static.input_text_buf[0..]))});
```
Pyrolistical, line 494: these `@as` can be removed if https://github.com/ziglang/zig/pull/19542 is accepted

samples/gui_test_wgpu/src/gui_test_wgpu.zig
```diff
@@ -478,12 +489,30 @@ fn update(demo: *DemoState) !void {
...
+            _ = zgui.text("length of Input text multiline {}", .{std.mem.len(@as([*:0]u8, static.input_text_multiline_buf[0..]))});
+            _ = zgui.inputTextWithHint("Input text with hint", .{
+                .hint = "Enter your name",
+                .buf = static.input_text_with_hint_buf[0..],
+            });
+            _ = zgui.text("length of Input text with hint {}", .{std.mem.len(@as([*:0]u8, static.input_text_with_hint_buf[0..]))});
+
+            zgui.separatorText("alloced input text");
+            _ = zgui.inputText("Input text alloced", .{ .buf = demo.alloced_input_text_buf });
+            _ = zgui.text("length of Input text alloced {}", .{std.mem.len(demo.alloced_input_text_buf.ptr)});
```
Pyrolistical, line 506: ditto for these `.ptr`

PR #543, opened 2024-04-05
[zgui][fix] bad mouse position with zgui+glfw and hidpi on macOS

OndraVoves:
I miss this little hack during update to new version.


hazeycode:
Seems weird that the vendored backend doesn't already do this. Should this fix be upstreamed?

PR #549, opened 2024-04-08
zgui: Expose DockBuilder API - partial coverage implemented and working

a-day-old-bagel:
Imgui's internal DockBuilder API has more functions than just these, which are not yet exposed, but these ones I've exposed are sufficient to use the API for basic examples (and for my purposes).

Let me know if there's anything that should be changed. I whipped this up and lightly tested it with my own code, as usual, but if more is needed for this to be merged, I can do more.

### DIFF
--- a/libs/zgui/src/gui.zig
+++ b/libs/zgui/src/gui.zig
@@ -881,7 +881,23 @@ pub const DockNodeFlags = packed struct(c_int) {
     no_resize: bool = false,
     auto_hide_tab_bar: bool = false,
     no_undocking: bool = false,
-    _padding: u24 = 0,
+    _padding_0: u2 = 0,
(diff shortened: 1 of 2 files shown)


a-day-old-bagel:
Oops, zig fmt --check didn't like how I split a function signature across multiple lines. I "fixed" that by making it less readable, so now the tests will hopefully pass.

hazeycode:
> Oops, zig fmt --check didn't like how I split a function signature across multiple lines. I "fixed" that by making it less readable, so now the tests will hopefully pass.

`,` after the last fn arg causes gets `zig fmt` to play along, see 382ff76d6862c014123aac59f64b0d6c8d56e1eb

PR #558, opened 2024-04-14
Automatically wire zig component deinit to ecs dtor

glitch-cake:


hazeycode:
Is this a breaking change?

Does this look good to you @Srekel ?

glitch-cake:
I'm not sure if this is breaking. I added a check to only perform the wiring for Structs which should fix the tests. I'm not sure what the correct/best thing to do for optionals of structs or pointers to structs etc, so I opted to not do anything.

Srekel:
Ah, so if you have a component with `deinit` it'll get called automatically? This shouldn't really affect any components unless you actually use this feature so I think it should be fine. I don't think this is something you'd want to do a lot but there might be a few cases. Soo I think it's good :) 👍

glitch-cake:
Thanks :) I'm in the process of adding a test right now, which is basically a minimal working example from my code base of how I'm using this.

I need to leave for a bit soon, but I'd like to get the test added before merging, and y'all can sanity check that the use-case perhaps makes sense, which then I would feel this is in a decent spot to merge if you all do too!

copygirl:
What about resources you pass into Flecs to have them be accessible that have a longer timespan than the component? Say if I have a `*Sprite` component on an entity to decide which sprite to render, but it's managed manually? I wouldn't want `deinit` to be called when removing the component is from the entity in this case.

Does it make sense for Flecs to call `deinit` if it doesn't even call `init` when being added to an entity?

Disclaimer: I don't use zflecs, I work on my own WIP Flecs library.

Srekel:
Maybe require the hook function to be called `dtor` - it doesn't match typical Zig naming but in this case it's probably good to be clear (in the component) that it's a flecs connection.

copygirl:
Mirroring the Flecs component hook names would be ideal and shouldn't interfere with function names typically used (unlike `deinit`). In this case, it would make sense to check for all of them, not just `dtor`, of course.

glitch-cake:
The use-case I have now which I've put in a test is a component with an ArrayList of messages. I want to deallocate the memory automatically when the component is destroyed.

I've renamed for it to use dtor instead of deinit, that makes sense.

If it's me I would skip doing ctor and others since I'm not using them and am not quite sure what the use case would look like. I could write the code, but I'm worried it might not be used/be totally cromulent.

PR #559, opened 2024-04-14
BREAKING: changed zd3d12 create*ShaderPipeline and MipmapGenerator to require shader bytecode

### LINKED ISSUE #556: zd3d12 lib MipmapGenerator depends on sample
[`MipmapGenerator`](https://github.com/zig-gamedev/zig-gamedev/blob/828a0786940976833b341deef791dc8303973f53/libs/zd3d12/src/zd3d12.zig#L1784) depends on a compiled compute shader under `content_dir ++ "shaders/generate_mipmaps.cs.cso"`

That file is build time compiled in each sample: https://github.com/zig-gamedev/zig-gamedev/blob/828a0786940976833b341deef791dc8303973f53/samples/simple_raytracer/build.zig#L88-L97

Second, `content_dir` is a sample concept that libs should not have.

Related to this I find it awkward a lib would directly load from the filesystem instead of be given the bytes. By allowing the caller to pass in the bytes, it can decide to either:
  1. load it from the filesystem
  2. [compile it at runtime](https://github.com/zig-gamedev/zig-gamedev/blob/main/libs/zwin32/src/d3dcompiler.zig)
  3. already have loaded via `@embedFile`

Pyrolistical:
Removes all `std.fs.openFileAbsolute` out of `zd3d12.zig`, but there still remain indirect file loading via `gctx.wic_factory.CreateDecoderFromFilename` and `dds_loading.zig`

Can be improved by extracting `csGenerateMipmaps` out of `samples/common/src/hlsl/common.hsls` and put it into zd3d12 with a shader compilation step template.

fixes #556

### DIFF
--- a/libs/zd3d12/src/zd3d12.zig
+++ b/libs/zd3d12/src/zd3d12.zig
@@ -975,69 +975,9 @@ pub const GraphicsContext = struct {
 
     pub fn createGraphicsShaderPipeline(
         gctx: *GraphicsContext,
-        arena: std.mem.Allocator,
         pso_desc: *d3d12.GRAPHICS_PIPELINE_STATE_DESC,
-        vs_cso_path: ?[]const u8,
-        ps_cso_path: ?[]const u8,
     ) PipelineHandle {
-        return createGraphicsShaderPipelineVsGsPs(gctx, arena, pso_desc, vs_cso_path, null, ps_cso_path);
-    }
-
-    pub fn createGraphicsShaderPipelineVsGsPs(
-        gctx: *GraphicsContext,
-        arena: std.mem.Allocator,
-        pso_desc: *d3d12.GRAPHICS_PIPELINE_STATE_DESC,
-        vs_cso_path: ?[]const u8,
-        gs_cso_path: ?[]const u8,
-        ps_cso_path: ?[]const u8,
-    ) PipelineHandle {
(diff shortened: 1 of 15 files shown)


hazeycode:
> Can be improved by extracting csGenerateMipmaps out of samples/common/src/hlsl/common.hsls and put it into zd3d12 with a shader compilation step template.

Agreed. Generally, I think we should be making samples/common smaller/non-existent.

Pyrolistical:
> Agreed. Generally, I think we should be making samples/common smaller/non-existent.

Ah, this explains a lot. `common` used to be a lib: https://github.com/zig-gamedev/zig-gamedev/commit/ff9bf9fe764736ac900b9a725730a23b60a8c333

hazeycode:
> @hazeycode do you know why all the windows d3d12 samples common imgui/vectormath instead of zgui/zmath?

Seems to be old code used by original Windows only version of zig-gamedev.

I started on a win32_dx12 zgui backend so that the common imgui can be dropped... I'll dig that out.

michal-z:
> > do you know why all the windows d3d12 samples common imgui/vectormath instead of zgui/zmath?
> 
> Seems to be old code used by original Windows only version of zig-gamedev.

Yes, exactly. D3D12 samples were written when `zmath` and `zgui` libs haven't existed yet.

PR #560, opened 2024-04-15
[zgui] Fix glfw-wgpu backend. Sry :)

OndraVoves:


hazeycode:
Seems to work fine with it calling the wrong thing. Weird. Anyway, thanks for following up with a fix.

PR #561, opened 2024-04-15
Removed minimal_d3d12 sample

Pyrolistical:
Related to https://github.com/zig-gamedev/zig-gamedev/pull/559#issuecomment-2057897450

If minimal_d3d12 where to be updated, it would become minimal_glfw_d3d12. So we can  just delete minimal_d3d12.


hazeycode:
minimal_d3d12 is a good reference for users who want to target win32 instead of glfw

PR #568, opened 2024-04-21
zphysics: ConvexHullShape: Some shape-exclusive methods exposed

a-day-old-bagel:
Since no other derived classes of Shape had yet been exposed, I did this in the way that I thought best fit the pattern laid out in the rest of zphysics.

If this isn't the way we want to go about exposing derived shape classes (in cases like these where the derived class has added functionality that isn't just an overridden virtual method from Shape), then I'd be happy to redo this in whatever way you might prefer.

As usual, I've tested these functions by using them all in my own project, but that's the extent of it.


hazeycode:
Looks good to me.
cc @michal-z

Issue #166, opened 2022-12-05
Wayland support


cmvanb:
Hi there,

I came across your repo by chance. I love the idea of using zig for games!

Unfortunately `zig build` fails for me with the following error:

```
./build.zig:149:60: error: expected type 'std.mem.Allocator', found 'std.zig.CrossTarget'
    const target = (std.zig.system.NativeTargetInfo.detect(cross) catch unreachable).target;
                                                           ^
/usr/lib/zig/std/mem/Allocator.zig:1:1: note: std.mem.Allocator declared here
//! The standard memory allocation interface.
^
/usr/lib/zig/std/zig/CrossTarget.zig:1:1: note: std.zig.CrossTarget declared here
//! Contains all the same data as `Target`, additionally introducing the concept of "the native target".
^
```

Is there a specific zig version I should be using? Or a dependency I may be missing?

I'm on Arch Linux with a wayland compositor.

michal-z:
Hi, thanks for the report. Project strictly follows latest changes in the Zig compiler so you will need to download latest master build from: https://ziglang.org/download/

michal-z:
One more thing - we don't currently support Wayland, only X11, sorry!

cmvanb:
Ahhhh busted. Oh well, thanks for the reply. Best of luck with the project, looks cool!

marler8997:
Just FYI, most X applications still work with Wayland compositors by going through Wayland.

michal-z:
Thanks for the info I didn't know about it. @cmvanb Could you try it out?

cmvanb:
I installed the latest version of zig.

```
zig version
0.11.0-dev.610+e7d28344f
```

Although `zig build` now completes, when I try to run any of the executables in `/zigout/bin` I get the following:

```
./triangle_wgpu
thread 248902 panic: reached unreachable code
libs/zglfw/src/zglfw.zig:45:5: 0xca6a88 in getRequiredInstanceExtensions (triangle_wgpu)
libs/zgpu/src/zgpu.zig:64:60: 0xc9e979 in impl (triangle_wgpu)
libs/zgpu/src/zgpu.zig:69:32: 0xc9db1f in create (triangle_wgpu)
samples/triangle_wgpu/src/triangle_wgpu.zig:57:49: 0xca010f in init (triangle_wgpu)
samples/triangle_wgpu/src/triangle_wgpu.zig:337:20: 0xca27b4 in main (triangle_wgpu)
[...]

PR #571, opened 2024-04-21
Wayland Support

hazeycode:
* Add wayland headers to system-sdk
* Compile glfw with wayland support by default
* Adds Wayland support to zgpu

Closes https://github.com/zig-gamedev/zig-gamedev/issues/166


Pyrolistical:
i think this broke minimal_glfw_d3d12.

looking into it.

PR #574, opened 2024-04-24
[zflecs]: Implement `comptime` for generating systems and `system_desc_t` from arbitrarly `fn`

gabrieldechichi:
This PR adds several comptime functions to `zflecs`, mainly allowing more egornomic system implementatinos, and system addition.

In short, it enables users to do this:

```zig
fn move_system(positions: []Position, velocities: []const Velocity) void {
    for (positions, velocities) |*p, v| {
        p.x += v.x;
        p.y += v.y;
    }
}

pub fn main() !void {
    ...
    ecs.ADD_SYSTEM(world, "move system", ecs.OnUpdate, move_system);
    ...
}
```

Which is more more ergonormic and less error prone than using `*iter_t` and defining queries manually on `system_desc_t`. It also marks the query term as readonly (`.In`) if the array is `const`.

This PR also supports passing `*iter_t` if the user wishes. For instance a user could have something like this:

```zig
fn move_system_with_it(it: *ecs.iter_t, positions: []Position, velocities: []const Velocity) void {
    const type_str = ecs.table_str(it.world, it.table).?;
    print("Move entities with [{s}]\n", .{type_str});
    defer ecs.os.free(type_str);

    for (positions, velocities) |*p, v| {
        p.x += v.x;
        p.y += v.y;
    }
}
```

Or define the queries manually like this:

```zig
fn move(it: *ecs.iter_t) callconv(.C) void {
    const p = ecs.field(it, Position, 1).?;
    const v = ecs.field(it, Velocity, 2).?;
    ...
}

pub fn main() !void {
    ...
        ecs.ADD_SYSTEM_WITH_FILTERS(world, "move system", ecs.OnUpdate, move, &.{
            .{ .id = ecs.id(Position) },
            .{ .id = ecs.id(Velocity), .inout = .In },
        });
    ...
}
```

### DIFF
--- a/libs/zflecs/README.md
+++ b/libs/zflecs/README.md
@@ -30,17 +30,22 @@ const Velocity = struct { x: f32, y: f32 };
 const Eats = struct {};
 const Apples = struct {};
 
-fn move(it: *ecs.iter_t) callconv(.C) void {
-    const p = ecs.field(it, Position, 1).?;
-    const v = ecs.field(it, Velocity, 2).?;
+fn move_system(positions: []Position, velocities: []const Velocity) void {
+    for (positions, velocities) |*p, v| {
+        p.x += v.x;
+        p.y += v.y;
+    }
+}
 
+//Optionally, systems can receive the components iterator (usually not necessary)
+fn move_system_with_it(it: *ecs.iter_t, positions: []Position, velocities: []const Velocity) void {
     const type_str = ecs.table_str(it.world, it.table).?;
-    std.debug.print("Move entities with [{s}]\n", .{type_str});
+    print("Move entities with [{s}]\n", .{type_str});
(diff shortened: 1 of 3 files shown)


hazeycode:
CC @Srekel, @btipling

hazeycode:
@gabrieldechichi thanks for the PR. Looks like this is failing checks because we enforce Zig's canonical formatting. Please run `zig fmt .` to auto format.

gabrieldechichi:
> @gabrieldechichi thanks for the PR. Looks like this is failing checks because we enforce Zig's canonical formatting. Please run `zig fmt .` to auto format.

Oh my bad! Just fixed it. I also ran the rest of the CI commands (for Windows) on `0.12.0-dev.2063+804cee3b9`, so hopefully CI should pass now.

PR #575, opened 2024-04-25
[zd3d12] fixed RenderTargetWriteMask default to be ALL

Pyrolistical:
According to https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_render_target_blend_desc

`RENDER_TARGET_BLEND_DESC.RenderTargetWriteMask` is the type of `COLOR_WRITE_ENABLE` and defaults to `ALL`.

This explains why there were a bunch of extra `pso_desc.BlendState.RenderTarget[0].RenderTargetWriteMask = 0xf` in our samples, while other d3d12 hello triangles used the default `BLEND_DESC`, which in turn used default `RENDER_TARGET_BLEND_DESC`


hazeycode:
Thanks

PR #579, opened 2024-04-28
Removed `init-default` in favor of default struct values

### LINKED ISSUE #576: [zd3d12] default values instead of `initDefault`?
As a follow up to #575

Why do we write:
```zig
pub const RENDER_TARGET_BLEND_DESC = extern struct {
    BlendEnable: BOOL,
    LogicOpEnable: BOOL,
    SrcBlend: BLEND,
    DestBlend: BLEND,
    BlendOp: BLEND_OP,
    SrcBlendAlpha: BLEND,
    DestBlendAlpha: BLEND,
    BlendOpAlpha: BLEND_OP,
    LogicOp: LOGIC_OP,
    RenderTargetWriteMask: COLOR_WRITE_ENABLE,

    pub fn initDefault() RENDER_TARGET_BLEND_DESC {
        var v = std.mem.zeroes(@This());
        v = .{
            .BlendEnable = FALSE,
            .LogicOpEnable = FALSE,
            .SrcBlend = .ONE,
            .DestBlend = .ZERO,
            .BlendOp = .ADD,
            .SrcBlendAlpha = .ONE,
            .DestBlendAlpha = .ZERO,
            .BlendOpAlpha = .ADD,
            .LogicOp = .NOOP,
            .RenderTargetWriteMask = COLOR_WRITE_ENABLE.ALL,
        };
        return v;
    }
};
```

when it can be

```zig
pub const RENDER_TARGET_BLEND_DESC = extern struct {
    BlendEnable: BOOL = FALSE,
    LogicOpEnable: BOOL =  FALSE,
    SrcBlend: BLEND = .ONE,
    DestBlend: BLEND = .ZERO,
    BlendOp: BLEND_OP = .ADD,
    SrcBlendAlpha: BLEND = .ONE,
    DestBlendAlpha: BLEND = .ZERO,
    BlendOpAlpha: BLEND_OP = .ADD,
    LogicOp: LOGIC_OP = .NOOP,
    RenderTargetWriteMask: COLOR_WRITE_ENABLE = COLOR_WRITE_ENABLE.ALL,
};
```

These default values some times are documented on the struct https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_render_target_blend_desc

but for other structs, they provide a helper https://learn.microsoft.com/en-us/windows/win32/direct3d12/cd3dx12-root-descriptor1

we can unify both in zig with default values.

vortexofdoom:
Would close #576

I removed the methods, although to avoid breakage they could be put in as wrappers around the new default constructors at first?

### DIFF
--- a/libs/zwin32/src/d3d12.zig
+++ b/libs/zwin32/src/d3d12.zig
@@ -895,80 +895,47 @@ pub const LOGIC_OP = enum(UINT) {
 };
 
 pub const RENDER_TARGET_BLEND_DESC = extern struct {
-    BlendEnable: BOOL,
-    LogicOpEnable: BOOL,
-    SrcBlend: BLEND,
-    DestBlend: BLEND,
-    BlendOp: BLEND_OP,
-    SrcBlendAlpha: BLEND,
-    DestBlendAlpha: BLEND,
-    BlendOpAlpha: BLEND_OP,
-    LogicOp: LOGIC_OP,
-    RenderTargetWriteMask: COLOR_WRITE_ENABLE,
+    BlendEnable: BOOL = FALSE,
+    LogicOpEnable: BOOL = FALSE,
+    SrcBlend: BLEND = .ONE,
+    DestBlend: BLEND = .ZERO,
+    BlendOp: BLEND_OP = .ADD,
+    SrcBlendAlpha: BLEND = .ONE,
+    DestBlendAlpha: BLEND = .ZERO,
+    BlendOpAlpha: BLEND_OP = .ADD,
+    LogicOp: LOGIC_OP = .NOOP,
+    RenderTargetWriteMask: COLOR_WRITE_ENABLE = COLOR_WRITE_ENABLE.ALL,
 
     pub fn initDefault() RENDER_TARGET_BLEND_DESC {
-        var v = std.mem.zeroes(@This());
-        v = .{
-            .BlendEnable = FALSE,
-            .LogicOpEnable = FALSE,
-            .SrcBlend = .ONE,
-            .DestBlend = .ZERO,
-            .BlendOp = .ADD,
-            .SrcBlendAlpha = .ONE,
-            .DestBlendAlpha = .ZERO,
-            .BlendOpAlpha = .ADD,
-            .LogicOp = .NOOP,
-            .RenderTargetWriteMask = COLOR_WRITE_ENABLE.ALL,
-        };
-        return v;
+        return .{};
     }
 };
 
 pub const BLEND_DESC = extern struct {
-    AlphaToCoverageEnable: BOOL,
-    IndependentBlendEnable: BOOL,
-    RenderTarget: [8]RENDER_TARGET_BLEND_DESC,
+    AlphaToCoverageEnable: BOOL = FALSE,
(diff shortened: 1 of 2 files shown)


copygirl:
Something the Zig standard library does, is keeping the functions, but having them just cause a `@compileError`, informing you of the change you need to make. Just wanted to mention that as another possible approach.

terraquad:
To avoid breakage, you could leave the `initDefault` functions there but they just return an "empty" instance of the struct. For example:
```zig
pub const RENDER_TARGET_BLEND_DESC = struct {
    BlendEnable: BOOL = FALSE,
    LogicOpEnable: BOOL = FALSE,
    SrcBlend: BLEND = .ONE,
    DestBlend: BLEND = .ZERO,
    BlendOp: BLEND_OP = .ADD,
    SrcBlendAlpha: BLEND = .ONE,
    DestBlendAlpha: BLEND = .ZERO,
    BlendOpAlpha: BLEND_OP =.ADD,
    LogicOp: LOGIC_OP = .NOOP,
    RenderTargetWriteMask: UINT8 = 0x0,

    // Leave this here for compatibility
    pub fn initDefault() @This() {
        // Returning an empty struct instance will do the same as the old initDefault function
        return .{};
    }
};
```

hazeycode:
> I removed the methods, although to avoid breakage they could be put in as wrappers around the new default constructors at first?

Yes, I think it's not worth breaking existing code with this change.

vortexofdoom:
I put the functions back in and they now all simply use the new default constructors.

hazeycode:
Thanks. Please rebase onto main and resolve conflicts.

vortexofdoom:
Done, found another internal usage of the old methods in `zd3d12.zig` that I removed, so at least within itself it should no longer use them.

vortexofdoom:
I ran `zig fmt` on it when CI failed, let me know if any other changes are needed.

libs/zwin32/src/d3d12.zig
```diff
@@ -895,80 +895,47 @@ pub const LOGIC_OP = enum(UINT) {
...
+    BlendEnable: BOOL = FALSE,
+    LogicOpEnable: BOOL = FALSE,
+    SrcBlend: BLEND = .ONE,
+    DestBlend: BLEND = .ZERO,
+    BlendOp: BLEND_OP = .ADD,
+    SrcBlendAlpha: BLEND = .ONE,
+    DestBlendAlpha: BLEND = .ZERO,
+    BlendOpAlpha: BLEND_OP = .ADD,
+    LogicOp: LOGIC_OP = .NOOP,
+    RenderTargetWriteMask: UINT8 = 0x0,
```
hazeycode, line 907: Looks like a conflict resolution error. The type here should be `COLOR_WRITE_ENABLE`

hazeycode (changes requested):

libs/zwin32/src/d3d12.zig
vortexofdoom: Not sure how those wires got crossed, fixed.

hazeycode: Thanks

PR #572, opened 2024-04-22
Up Zig to 0.12.0

hazeycode:

PR #584, opened 2024-04-28
Make zopengl compatible with Zig 0.12.0

terraquad:
The Linkage enum values are written in lowercase in Zig 0.12.0
Without this change, there is a compile error.


hazeycode:
This is already changed on our unstable branch and also on 0.12.0 branch along with other 0.12.0 changes here: https://github.com/zig-gamedev/zig-gamedev/pull/572

Note that we will be skipping 0.12.0 on the main branch since there are issues upstream that have been fixed more recently.

PR #586, opened 2024-04-30
Add zopenvr

Pyrolistical:
Implemented [OpenVR SDK](https://github.com/ValveSoftware/openvr) as `zopenvr` lib. `zopenvr` is a wip. See its readme for implemented interfaces.

Add two samples:
1. `openvr_test`: a zgui app that allows one to manually make OpenVR api calls and see results
2. `simple_openvr`: a port of a OpenVR sample

OpenVR SDK is BSD licensed, so there should be no concern porting the OpenVR sample source or including the assets.

I had to make some changes to zd3d12, but I kept those changes in its own commit 3401400863790076c92ed9221adfd4fd2d2ca335

This has only been tested on windows. In theory it should work in linux as well.


Pyrolistical:
This is a massive PR. I understand if you take your time.

Pyrolistical:
let me simplify zopenvr by moving all the public structs into common. this will flatten the namespace

hazeycode:
To get the sample apps to launch I needed to add
```zig
// We need to export below symbols for DirectX 12 Agility SDK.
pub export const D3D12SDKVersion: u32 = 610;
pub export const D3D12SDKPath: [*:0]const u8 = ".\\d3d12\\";
```
to  openvr_test.zig and simple_openvr.zig

and
```zig
    // This is needed to export symbols from an .exe file.
    // We export D3D12SDKVersion and D3D12SDKPath symbols which
    // is required by DirectX 12 Agility SDK.
    exe.rdynamic = true;
```
to each build.zig

Pyrolistical:
> To get the sample apps to launch I needed to add

fixed

Pyrolistical:
If this is accepted, would like to support wgpu. The tricky part is the openvr compositor only supports directx, vulkan and opengl textures. While dawn does implement those backends, afaik there isn't a way to to get the native texture given a wgpu texture.

An escape hatch like that would be generally useful.

hazeycode:
Nice contribution. Thanks!

PR #589, opened 2024-05-07
[zd3d12] added helpers for CreateShaderResourceView, CreateUnorderedAccessView, SetGraphicsRootDescriptorTable and SetComputeRootDescriptorTable

Pyrolistical:


samples/rasterization/src/rasterization.zig
```diff
@@ -341,11 +334,6 @@ fn init(allocator: std.mem.Allocator) !DemoState {
         ) catch |err| hrPanic(err),
     };
 
-    for (mesh_textures) |texture| {
```
Pyrolistical, line 344: pretty sure this is deadcode that all it did was take up memory

hazeycode: @michal-z can you confirm if it's ok to remove this?

michal-z:
This code is required (a bit misleading though). We are creating SRVs using bindless approach at the beginning of the heap. So, the resources are then accessed in the shader like this:
```hlsl
    Texture2D srv_ao = ResourceDescriptorHeap[0];
    Texture2D srv_base_color = ResourceDescriptorHeap[1];
    Texture2D srv_metallic_roughness = ResourceDescriptorHeap[2];
    Texture2D srv_normal = ResourceDescriptorHeap[3];
```

Pyrolistical:
OK. Let me fix that

Pyrolistical:
Fixed but bindless is indeed very misleading. It seems only useful if generating HLSL as having a human keeping the HLSL in sync with the code usage of heap is error prone.

michal-z:
> Fixed but bindless is indeed very misleading. It seems only useful if generating HLSL as having a human keeping the HLSL in sync with the code usage of heap is error prone.

In general bindless is very powerful and provides minimum CPU overhead. I was just lazy in this sample code and I assumed that we have only 4 SRVs and the beginning of the heap.

You can take a look at my new project which uses bindless in much better way (simple, fast and robust):

Header file which is used by both Zig and HLSL code and defines resource indices:
https://github.com/michal-z/zig-d3d12-starter/blob/main/src/cpu_gpu_shared.h#L1

In HLSL, resources are used like this:
https://github.com/michal-z/zig-d3d12-starter/blob/main/src/main.hlsl#L40

In Zig, SRVs are defined like this:
https://github.com/michal-z/zig-d3d12-starter/blob/main/src/gen_mesh.zig#L607

Issue #596, opened 2024-05-15
LazyPath.path is deprecated


Pyrolistical:
In the current pinned version, LazyPath.path is deprecated:
https://github.com/ziglang/zig/blob/3648d7df19642b6902b3f75538a826a322211673/lib/std/Build.zig#L2135-L2136

But it is used throughout the builds:
https://github.com/search?q=repo%3Azig-gamedev%2Fzig-gamedev+.path+language%3AZig++path%3A**%2Fbuild.zig&type=code

Pyrolistical:
...and its gone:
https://github.com/ziglang/zig/commit/dee9f82f69db0d034251b844e0bc4083a1b25fdd

Pyrolistical:
I'll get a PR up

PR #600, opened 2024-05-18
Remove deprecated LazyPath.path

Pyrolistical:
Fixes https://github.com/zig-gamedev/zig-gamedev/issues/596


hazeycode:
Nice. Thanks!

PR #603, opened 2024-05-22
Add custom zpool handle methods

tibimunteanu:
This PR proposes a small change that adds the ability to define custom methods in the handle namespace of `zpool`.
While working with `zpool`, I found myself wanting to have a cleaner and shorter way of working with the handles.
This is subjective and I am not saying that this is better or that this is how things should be.

I wanted to turn code like:
`texture_pool.getColumnPtr(handle, .texture);`
into
`handle.getPtr();`
or turn code like:
```
texture_pool.requireLiveHandle(handle);
texture_pool.remove(handle);
```
into
`handle.release();`

For more details, have a look at the last test `Pool with custom handle methods` in the `pool.zig` file, where I give a more comprehensive example of a `TextureSystem` that ends up with a clean api thanks to the proposed change.

I'm sure that this could be achieved in a cleaner and more non-breaking way via some zig comptime magic or if creating a Pool takes a single struct parameter with the THandleMethods initialized to an empty struct so you only specify it if you want to add custom methods to the handle.

I am open to learning if all this is not a good idea and why.

### DIFF
--- a/libs/zgpu/src/zgpu.zig
+++ b/libs/zgpu/src/zgpu.zig
@@ -1392,7 +1392,7 @@ const PipelineLayoutPool = ResourcePool(PipelineLayoutInfo, wgpu.PipelineLayout)
 
 fn ResourcePool(comptime Info: type, comptime Resource: type) type {
     const zpool = @import("zpool");
-    const Pool = zpool.Pool(16, 16, Resource, struct { info: Info });
+    const Pool = zpool.Pool(16, 16, Resource, struct { info: Info }, struct {});
 
     return struct {
         const Self = @This();
--- a/libs/zpool/src/handle.zig
+++ b/libs/zpool/src/handle.zig
@@ -45,6 +45,7 @@ pub fn Handle(
     comptime index_bits: u8,
     comptime cycle_bits: u8,
     comptime TResource: type,
+    comptime THandleMethods: type,
 ) type {
     if (index_bits == 0) @compileError("index_bits must be greater than 0");
     if (cycle_bits == 0) @compileError("cycle_bits must be greater than 0");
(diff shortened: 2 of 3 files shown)


hazeycode:
It's not clear to me that this is any better than just writing something like:
```zig
const TextureSystem = struct {
    ...
    pub fn acquire(handle: Handle) !void {
        try pool.requireLiveHandle(handle);
        const reference_count = pool.getColumnPtrAssumeLive(self, .reference_count);
        reference_count.* +|= 1;
    }
}
```
which may be called like:
```zig
    try texture_system.aquire(handle);
```

Does this convention meet your needs?

tibimunteanu:
I am not saying that the current way doesn't work, but just that the api can be more concise and cleaner if you can use the handle more like a reference rather than explicitly doing everything through the `texture_system`. Why does the user of the api need to know that there is a system when he only wants to work with a texture?

```zig
if (handle.getPtrIfExists()) |texture_1| {
    std.log.info("{s}", .{ texture_1.name });
}

if (handle.exists()) {
    try handle.acquire();
}

try handle.release();
```
vs
```zig
if (texture_system.getColumnPtrIfLive(handle, .texture)) |texture_1| {
    std.log.info("{s}", .{ texture_1.name });
}

if (texture_system.isLiveHandle(handle)) {
    texture_system.acquireTexture(handle);
}

texture_system.releaseTexture(handle);
```

hazeycode:
I only included TextureSystem because your example code uses that. It could just as easily be written as a top level function.

This change proposes an api that encourages an object orientated style that serves little to no practical benefit at the cost of increased api surface, obfuscation of ownership and increased compile times.

I am open to reconsidering my position on this if there is a use case presented where the benefit is obvious.

PR #605, opened 2024-05-23
Added live coding sample

Pyrolistical:
https://github.com/zig-gamedev/zig-gamedev/assets/463642/32e91e74-1cea-43ce-8f2c-ae3767015a52

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -95,6 +95,7 @@ const samples_windows = struct {
     pub const audio_playback_test = @import("samples/audio_playback_test/build.zig");
     pub const directml_convolution_test = @import("samples/directml_convolution_test/build.zig");
     pub const vector_graphics_test = @import("samples/vector_graphics_test/build.zig");
+    pub const live_editing = @import("samples/live_editing/build.zig");
 };
 
 const samples_windows_linux = struct {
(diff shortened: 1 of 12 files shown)


Srekel:
Cool!

Pyrolistical:
Inspired by the first demo of https://vimeo.com/906418692

hazeycode:
This is cool. But I don't think we should merge it. I'd like to try and lean on Zig's incoming hot code reload features and support as many samples as possible. The asset reloading portion of this PR would still be applicable to such a set up.

Pyrolistical:
I agree. This was an experiment to see what was possible before https://github.com/ziglang/zig/pull/20580

Srekel:
Would this be useful as a sample for non-zig hot reloadable stuff? For example someone might have part of their engine in a DLL that could be written in C. Or if you have shaders or even textures or other content, listening for changes on those would be very useful. And personally I don't want to put that through the Zig build system.

hazeycode:
I think patching in C or C++ portions will be serviceable with the Zig build system. But it perhaps does make sense to have an example of a hot reloading asset pipeline as a standalone sample app rather than trying to factor it into the existing ones. I do think we should stick to purely Zig driven code reloading in this repo though; there are bunch of examples of the shared lib approach to hot reloading game code elsewhere.

PR #606, opened 2024-05-24
zgui: remove a redundant installArtifact call, add `setWindowFocus`

kcbanner:
This extra call was causing `artifact("imgui")` calls to fail with an ambiguous artifact error (since "imgui" was being exposed twice).

Also added `setWindowFocus`.


hazeycode:
Thanks!

PR #614, opened 2024-06-01
Update zsdl example

terraquad:
It seems like the build methods from zsdl don't need the `zsdl_path` parameter anymore. Otherwise Zig throws an error.


hazeycode:
Thanks!

PR #619, opened 2024-06-11
Migrate system-sdk to it's own repo

hazeycode:
This [upstream issue](https://github.com/ziglang/zig/issues/20256) seems to be a particular problem for `system-sdk`, I think because it is big and more likely than smaller packages to fail in the fashion described by linked issues.

We are already using the package manager to fetch from GitHub archive URLs (zgpu and zsdl have such dependencies) and this is generally not a problem, at least on my own internet connection. Maybe all of our libs except for `system-sdk` will evade the underlying problem for all practical purposes (because they are smaller?) .


hazeycode:
Trying a smaller lib here: https://github.com/zig-gamedev/zig-gamedev/pull/667

hazeycode:
Superseded by https://github.com/zig-gamedev/zig-gamedev/pull/668

PR #623, opened 2024-06-14
[samples] refactored to use std.time constants

Pyrolistical:
added frame rate limiter to physics_test_wgpu as physics was frame rate based


hazeycode:
Nice one

PR #624, opened 2024-06-14
[zpix] Fixed zpix to work with latest release

Pyrolistical:
In the latest release of PIX, the installation path is `C:\Program Files\Microsoft PIX\2405.15.002-OneBranch_release`. The previous code breaks as it assumed the directory would be parseable as a float, IE. `2405.15`.

Changed `zpix` to require the `path` build option instead of trying to auto detect the latest version.


samples/triangle/build.zig
```diff
@@ -37,6 +37,12 @@ pub fn build(b: *std.Build, options: Options) *std.Build.Step.Compile {
     const zd3d12_module = zd3d12.module("root");
     exe.root_module.addImport("zd3d12", zd3d12_module);
 
+    const zpix = b.dependency("zpix", .{
+        .enable = false,
```
Pyrolistical, line 41: Intentionally disabled so the sample is runnable without PIX installed

PR #634, opened 2024-07-10
zgui: ImGuizmo included as optional feature, sample of usage added to monolith

a-day-old-bagel:
I added ImGuizmo to zgui [as per a discussion from earlier](https://github.com/zig-gamedev/zig-gamedev/discussions/611)

I tried to follow the pattern laid out by the addition of implot, but let me know if anything needs to be done differently here.

I added ImGuizmo functionality to the monolith sample (since the whole point of monolith was to offer an example of integrating multiple parts of zig-gamedev, and to allow demoing new things without creating whole new samples).

That inclusion in the monolith sample is what I'm passing off as "testing," but, again, let me know if you want more tests written or anything else changed.

image

### DIFF
--- a/libs/zgui/build.zig
+++ b/libs/zgui/build.zig
@@ -25,6 +25,11 @@ pub fn build(b: *std.Build) void {
             "with_implot",
             "Build with bundled implot source",
         ) orelse true,
+        .with_gizmo = b.option(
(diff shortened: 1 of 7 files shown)


hazeycode:
Thanks. This is great!

PR #638, opened 2024-07-15
[zmath] Add & use expectVecApproxEqAbs and expectVecEqual.

dmurph:
Adding vec-based testing functions that print the failures much better. These use inner expect calls, which will throw an error on the specific mismatch instead of an opaque failure just on the line being false.

### DIFF
--- a/libs/zmath/src/zmath.zig
+++ b/libs/zmath/src/zmath.zig
@@ -353,17 +353,17 @@ test "zmath.load" {
     var ptr = &a;
     var i: u32 = 0;
     const v0 = load(a[i..], F32x4, 2);
-    try expect(approxEqAbs(v0, F32x4{ 1.0, 2.0, 0.0, 0.0 }, 0.0));
+    try expectVecEqual(v0, F32x4{ 1.0, 2.0, 0.0, 0.0 });
     i += 2;
     const v1 = load(a[i .. i + 2], F32x4, 2);
-    try expect(approxEqAbs(v1, F32x4{ 3.0, 4.0, 0.0, 0.0 }, 0.0));
+    try expectVecEqual(v1, F32x4{ 3.0, 4.0, 0.0, 0.0 });
     const v2 = load(a[5..7], F32x4, 2);
-    try expect(approxEqAbs(v2, F32x4{ 6.0, 7.0, 0.0, 0.0 }, 0.0));
+    try expectVecEqual(v2, F32x4{ 6.0, 7.0, 0.0, 0.0 });
     const v3 = load(ptr[1..], F32x4, 2);
-    try expect(approxEqAbs(v3, F32x4{ 2.0, 3.0, 0.0, 0.0 }, 0.0));
+    try expectVecEqual(v3, F32x4{ 2.0, 3.0, 0.0, 0.0 });
     i += 1;
     const v4 = load(ptr[i .. i + 2], F32x4, 2);
-    try expect(approxEqAbs(v4, F32x4{ 4.0, 5.0, 0.0, 0.0 }, 0.0));
+    try expectVecEqual(v4, F32x4{ 4.0, 5.0, 0.0, 0.0 });
 }
 
 pub fn store(mem: []f32, v: anytype, comptime len: u32) void {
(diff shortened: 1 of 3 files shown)


libs/zmath/src/zmath.zig
```diff
@@ -4490,18 +4490,32 @@ fn floatToIntAndBack(v: anytype) @TypeOf(v) {
...
+}
+
+pub fn expectVecApproxEqAbs(expected: anytype, actual: anytype, eps: f32) !void {
+    const T = @TypeOf(expected, actual);
+    inline for (0..veclen(T)) |i| {
+        try std.testing.expectApproxEqAbs(expected[i], actual[i], eps);
+    }
+}
+
 pub fn approxEqAbs(v0: anytype, v1: anytype, eps: f32) bool {
```
dmurph, line 4519: note: I left this in case of dependencies & to not break people, happy to remove if desired.

hazeycode:
> will throw an error on the specific mismatch instead of an opaque failure just on the line being false.

This is not what I'm seeing. The output is quite generic and I think adds no useful information. Am I missing something?

Example (without this change):
```
test
└─ run zmath-tests 79/84 passed, 1 failed, 4 skipped
error: 'zmath.test.zmath.round' failed: /Users/chris/zig/0.13.0-dev.351+64ef45eb0/files/lib/std/testing.zig:540:14: 0x102138e73 in expect (zmath-tests)
    if (!ok) return error.TestUnexpectedResult;
             ^
/Users/chris/gamedev/zig-gamedev/libs/zmath/src/zmath.zig:955:5: 0x102140f93 in test.zmath.round (zmath-tests)
    try expect(approxEqAbs(v, f32x4(-1_388_610.0, 1_388_910.0, 1_388_109.0, 2_388_610.0), 0.0));
    ^
error: while executing test 'util.test.zmath.util.mat.right', the following test command failed:
/Users/chris/gamedev/zig-gamedev/libs/zmath/.zig-cache/o/9ce519e742c63ef799a10003aafec9b8/zmath-tests --listen=- 
```

Example (with this change):
```
test
└─ run zmath-tests 79/84 passed, 1 failed, 4 skipped
error: 'zmath.test.zmath.round' failed: expected 1.38861e6, found 1.38891e6
/Users/chris/zig/0.13.0-dev.351+64ef45eb0/files/lib/std/testing.zig:93:17: 0x10493cec3 in expectEqualInner__anon_1354 (zmath-tests)
                return error.TestExpectedEqual;
                ^
/Users/chris/gamedev/zig-gamedev/libs/zmath/src/zmath.zig:4508:9: 0x10493d143 in expectVecEqual__anon_1304 (zmath-tests)
        try std.testing.expectEqual(expected[i], actual[i]);
        ^
/Users/chris/gamedev/zig-gamedev/libs/zmath/src/zmath.zig:955:5: 0x10494587f in test.zmath.round (zmath-tests)
    try expectVecEqual(v, f32x4(-1_388_610.0, 1_388_910.0, 1_388_109.0, 2_388_610.0));
    ^
error: while executing test 'util.test.zmath.util.mat.right', the following test command failed:
/Users/chris/gamedev/zig-gamedev/libs/zmath/.zig-cache/o/a7b0b4f22e1898eb4eadbbd6f5506c76/zmath-tests --listen=- 
```

dmurph:
See the top line:
`error: 'zmath.test.zmath.round' failed: expected 1.38861e6, found 1.38891e6`

This now prints the exact mismatch.

libs/zmath/src/zmath.zig
hazeycode: Yes let's just remove it please. We can mark the PR as a breaking change.

hazeycode:
> An improvement here would be to save all errors for the given vec, and then return them all somehow so all mismatches are printed. Not sure how to do that, happy to make that change though if you have guidance.

I don't think it's worth the effort. But if someone figures out how to do it cleanly, it'd be nice.

libs/zmath/src/zmath.zig
dmurph: Done. I had to keep this for a negation check, so I renamed it to match the testing methods.

dmurph:
I removed approxEqAbs, but I still needed it for negation checks, so I renamed it to vecApproxEqAbs....

But at that point I think we should just keep the original to not break other potential users. What do you think?

hazeycode:
I agree. If we are keeping it, not worth renaming.

dmurph:
Done

hazeycode:
Thanks!

PR #645, opened 2024-07-24
zopengl: implement deleteProgram and deleteBuffer functions in wrapper

MichaelByrneAU:
Noticed that these were missing from the current wrapper implementation. Let me know if there was a reason for this that I missed!


hazeycode:
Hi and thanks! The wrapper is very much a work-in-progress. Only a small subset of the API has been wrapped so far, in order to prove it out.

PR #646, opened 2024-07-25
Support zigglgen

guidoschmidt:
I was trying to use zglfw in combination with [zigglgen](https://github.com/castholm/zigglgen). Turned out to get it working, it only needed a more specific type of `GlProc` in zglfw. Then also a change in `zopengl` was needed to be able to build and run the zglfw/gl samples in zig-gamedev.

Not quite sure about potential flaws (or potential issues) of this type change, so I'm opening a PR here to discuss.

### DIFF
--- a/libs/zopengl/src/zopengl.zig
+++ b/libs/zopengl/src/zopengl.zig
@@ -12,7 +12,7 @@ test {
 pub const bindings = @import("bindings.zig");
 pub const wrapper = @import("wrapper.zig").Wrap(bindings);
 
-pub const LoaderFn = *const fn ([:0]const u8) ?*const anyopaque;
+pub const LoaderFn = *const fn ([*:0]const u8) callconv(.C) ?*const fn () callconv(.C) void;
 
 pub const Extension = enum {
     KHR_debug,
--- a/libs/zglfw/src/zglfw.zig
+++ b/libs/zglfw/src/zglfw.zig
@@ -142,10 +142,12 @@ extern fn glfwGetCurrentContext() *Window;
 pub const swapInterval = glfwSwapInterval;
 extern fn glfwSwapInterval(interval: i32) void;
 
-pub const GlProc = *const anyopaque;
+pub const GlProc = *const fn () callconv(if (builtin.os.tag == .windows and builtin.cpu.arch == .x86) .Stdcall else .C) void;
 
-pub fn getProcAddress(procname: [:0]const u8) ?GlProc {
-    return glfwGetProcAddress(procname);
+
+pub fn getProcAddress(proc_name: [*:0]const u8) callconv(.C) ?GlProc {
+    if (glfwGetProcAddress(proc_name)) |proc_address| return proc_address;
+    return null;
 }
 extern fn glfwGetProcAddress(procname: [*:0]const u8) ?GlProc;
 


hazeycode:
Hi. What version of Zig are you using?

hazeycode:
Is it the calling convention that is the issue here?
I think we want to keep glfw and sdl gl proc type and a type erased pointer and do the appropriate cast on the user side. 
I am confused because zigglgen docs say "loader" is duck-typed (https://github.com/castholm/zigglgen#proctableinit). Can you post your code?

libs/zopengl/src/zopengl.zig
```diff
@@ -11,7 +11,7 @@ comptime {
 pub const bindings = @import("bindings.zig");
 pub const wrapper = @import("wrapper.zig").Wrap(bindings);
 
-pub const LoaderFn = *const fn ([:0]const u8) ?*const anyopaque;
+pub const LoaderFn = *const fn ([*:0]const u8) callconv(.C) ?*const fn () callconv(.C) void;
```
hazeycode, line 14: Why callconv(.C) ?*const fn () callconv(.C) void;`? This is not correct for many gl procs.

hazeycode (changes requested):
This seems to specific. is there another way to make this work with C calling convention?

guidoschmidt:
Maybe I was doing something wrong in the first place. The inital error I get (without my pointer type adjustments) is the following:

```shell
run
└─ run sketchbook.new.zig
   └─ zig build-exe sketchbook.new.zig Debug native 1 errors
.zig-cache/o/dc3ef1dc485fa6f0b2b00c181fe12794/gl.zig:2609:38: error: expected type '[:0]const u8', found '[*:0]const u8'
            return @as(?PROC, loader(@as([*:0]const u8, prefixed_name)));
                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
libs/zig-gamedev/libs/zglfw/src/zglfw.zig:152:33: note: parameter type declared here
pub fn getProcAddress(procname: [:0]const u8) ?GlProc {
                                ^~~~~~~~~~~~
referenced by:
    initCommand__anon_6823: .zig-cache/o/dc3ef1dc485fa6f0b2b00c181fe12794/gl.zig:2595:27
    init__anon_5326: .zig-cache/o/dc3ef1dc485fa6f0b2b00c181fe12794/gl.zig:2584:66
    remaining reference traces hidden; use '-freference-trace' to see all reference traces
error: the following command failed with 1 compilation errors:
/Users/gs/.zvm/0.13.0/zig build-exe /Users/gs/git/zig/sketchbook.new.zig/.zig-cache/o/f66cf3cf5e8329392f417fac17dea894/libglfw.a /Users/gs/git/zig/sketchbook.new.zig/.zig-cache/o/2d5f52bd28006b02cc6022b682541325/libzstbi.a -ODebug -I /Users/gs/git/zig/sketchbook.new.zig/.zig-cache/o/544790c738d38eb183fa637997b2e9c3 -I /Users/gs/git/zig/sketchbook.new.zig/.zig-cache/o/544790c738d38eb183fa637997b2e9c3 --dep zosc --dep zreload --dep zglfw --dep zstbi --dep gl -Mroot=/Users/gs/git/zig/sketchbook.new.zig/src/main.zig --dep network -Mzosc=/Users/gs/git/zig/sketchbook.new.zig/libs/zosc/src/lib.zig -Mzreload=/Users/gs/git/zig/sketchbook.new.zig/libs/zreload/src/lib.zig -I /Users/gs/git/zig/sketchbook.new.zig/libs/zig-gamedev/libs/zglfw/libs/glfw/include --dep zglfw_options -Mzglfw=/Users/gs/git/zig/sketchbook.new.zig/libs/zig-gamedev/libs/zglfw/src/zglfw.zig -Mzstbi=/Users/gs/git/zig/sketchbook.new.zig/libs/zig-gamedev/libs/zstbi/src/zstbi.zig -Mgl=/Users/gs/git/zig/sketchbook.new.zig/.zig-cache/o/dc3ef1dc485fa6f0b2b00c181fe12794/gl.zig -lobjc -Mnetwork=/Users/gs/.cache/zig/p/122080b2675fa7fdd7410af5f62942dec99e0ea766827c8c6506e6672b575fb0a50e/network.zig -Mzglfw_options=/Users/gs/git/zig/sketchbook.new.zig/.zig-cache/c/7c9255a4544989acb78aec85001f9df4/options.zig -framework IOKit -framework CoreFoundation -framework Metal -framework AppKit -framework CoreServices -framework CoreGraphics -framework Foundation -lc --cache-dir /Users/gs/git/zig/sketchbook.new.zig/.zig-cache --global-cache-dir /Users/gs/.cache/zig --name sketchbook.new.zig --listen=-
Build Summary: 7/12 steps succeeded; 1 failed (disable with --summary none)
run transitive failure
└─ run sketchbook.new.zig transitive failure
   ├─ zig build-exe sketchbook.new.zig Debug native 1 errors
   └─ install transitive failure
      └─ install sketchbook.new.zig transitive failure
         └─ zig build-exe sketchbook.new.zig Debug native (+6 more reused dependencies)
error: the following build command failed with exit code 1:
/Users/gs/git/zig/sketchbook.new.zig/.zig-cache/o/e8563506677f0f7e8e3372574b214b8e/build /Users/gs/.zvm/0.13.0/zig /Users/gs/git/zig/sketchbook.new.zig /Users/gs/git/zig/sketchbook.new.zig/.zig-cache /Users/gs/.cache/zig --seed 0x91f423f5 -Z05c453d65999a7fe run
```

I'm still learning pointers, align and callconv, so maybe it's just a lack of understanding on my end 🤔
So if we keep the proc table point type erased, which makes sense to me, most likely zigglglen needs to be adjusted in some way. I'm going to try that, thanks already for having a look 🙏🏽

guidoschmidt:
Working on this here: https://github.com/guidoschmidt/zigglgen/tree/fix/adjust-for-zig-gamedev-zglfw

castholm:
zigglgen author here.

@guidoschmidt
The reason you can't just pass zglfw's `getProcAddress` to zigglgen's `ProcTable.init` is because `getProcAddress` takes a `[:0]const u8` (0-terminated slice) while `ProcTable.init` expects a function that takes a `[*:0]const u8` (0-terminated many-pointer).

Instead of modifying zigglgen, you can work around this by providing a function that wraps zglfw's `getProcAddress` and uses [`std.mem.span`](https://ziglang.org/documentation/master/std/#std.mem.span) to convert the many-pointers into slices:

```zig
const std = @import("std");
const glfw = @import("zglfw");

fn getProcAddress(prefixed_name: [*:0]const u8) ?gl.PROC {
    return zglfw.getProcAddress(std.mem.span(prefixed_name));
}

pub fn main() !void {
    // ...

    if (!gl_procs.init(fixedGetProcAddress)) return error.InitFailed;

    // ...
}
```

@hazeycode
However, I would argue that it would probably be for the best if the signature for zglfw's `getProcAddress` was changed to `pub fn getProcAddress(procname: [*:0]const u8) ?GlProc`, since all that function does is forward the proc name to `glfwGetProcAddress` which indeed takes a `[*:0]const u8`. Requiring the proc name to be a slice arbitrarily restricts users for no reason and forces unnecessary conversions from many-pointers to spans.

The type of `GlProc` shouldn't need to be changed in practice, but if you want to be pedantic and theoretically correct, the OpenGL specs define the opaque OpenGL function pointers as `*const fn () callconv(if (builtin.os.tag == .windows and builtin.cpu.arch == .x86) .Stdcall else .C) void`. Since the function pointer needs to be casted anyway, the main important part is the alignment, so if you prefer returning an `anyopaque` pointer, it should be aligned properly, i.e. `*align(@alignOf(fn () callconv(if (builtin.os.tag == .windows and builtin.cpu.arch == .x86) .Stdcall else .C) void)) const anyopaque`.

(The calling convention stuff is based on the OpenGL `APIENTRY` calling convention macro: https://github.com/KhronosGroup/EGL-Registry/blob/7db3005d4c2cb439f129a0adc931f3274f9019e6/api/KHR/khrplatform.h#L122-L127)

guidoschmidt:
@castholm wow, nice thanks so much for stepping in 👍🏽

guidoschmidt:
Ok, I just tried. @castholm your proposed solution works, but I also had to add `@alignCast`:

```

fn getProcAddress(prefixed_name: [*:0]const u8) ?gl.PROC {
    return @alignCast(zglfw.getProcAddress(std.mem.span(prefixed_name)));
}
```

I did create an example of using *zigglen* in combination with *zigglgen*, see https://github.com/castholm/zigglgen/pull/6 

@hazeycode would it be benefitial to also add an example here in zig-gamedev? Or should we keep the examples free from third party libraries?

hazeycode:
@castholm thanks for pointing out the issue of alignment. I will consider your recommendation.
> However, I would argue that it would probably be for the best if the signature for zglfw's getProcAddress was changed to pub fn getProcAddress(procname: [*:0]const u8) ?GlProc, since all that function does is forward the proc name to glfwGetProcAddress which indeed takes a [*:0]const u8. Requiring the proc name to be a slice arbitrarily restricts users for no reason and forces unnecessary conversions from many-pointers to spans.

Agreed. I have made this change here: https://github.com/zig-gamedev/zig-gamedev/pull/646

@guidoschmidt this change should mean you won't need to do the workaround.

would it be benefitial to also add an example here in zig-gamedev? Or should we keep the examples free from third party libraries?

PR #647, opened 2024-07-25
add build as shared option option

DDosha:
This is a change i did for my usage and thought i should make a PR in case people needed it.


hazeycode:
Thanks!

PR #230, opened 2023-02-27
Support building dependencies as a shared library

kcbanner:
- Add build option to generated a shared library
- Add gui.initNoContext / gui.deinitNoContext
- Update ConfigFlags to be a packed struct
- Add gui.menuItemPtr / zguiMenuItemPtr
- Add extern "C" and ZGUI_API define to zgui.cpp

PR #655, opened 2024-07-30
zphysics: shared library support

kcbanner:
- Add support for building `joltc` as a shared library
- Add support for setting `Trace` and `AssertFailed` functions
- Add support for exporting / importing the global state, in order to support dll reloading
- Remove `.kfreebsd` as it's no longer present in std

The DLL reloading use case is somewhat niche, so if that isn't something that makes sense here I'm happy to maintain that part in my own fork.

In my project I use a hot-reload approach where the game dll gets reloaded after code changes, so any global state / vtables need to be updated. For context, I made similar changes here: https://github.com/zig-gamedev/zig-gamedev/pull/230

### DIFF
--- a/libs/zglfw/src/zglfw.zig
+++ b/libs/zglfw/src/zglfw.zig
@@ -1048,7 +1048,6 @@ fn _isLinuxDesktopLike() bool {
     return switch (builtin.target.os.tag) {
         .linux,
         .freebsd,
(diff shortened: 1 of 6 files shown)


hazeycode:
Thanks! Building joltc as a shared lib is something I think we should support.

PR #656, opened 2024-07-30
zgui: expose ListClipper API

cubedtear:
Exposes ImGui's ListClipper to have performant lists/tables with lots of items.

I tested it with my own code, and it works as expected.

Let me know if there is anything missing or that should be changed.


Srekel:
Oh nice, I was actually just going to start using this and was hoping it was bound already! :)

Issue #649, opened 2024-07-26
Update zflecs to flecs 4


giusdp:
Hey folks.

Flecs had very recently a new major version, perhaps this issue can be used to keep track for the update of the relative zflecs.
If possible I could try my hand at the update, although I'm pretty new to zig and I might need some guidance.

I think the major change is the removal of `filter`, as it was merged with `query`. I saw zflecs already has several tests in place, so there is a way to check if things are still working after the migration to flecs 4.

hazeycode:
Completed by https://github.com/zig-gamedev/zig-gamedev/pull/673

PR #657, opened 2024-08-01
Update zflecs to 4.0.0

giusdp:
Just did a first pass to update the bindings for the new zflecs version. The tests not using `ecs.progress` pass, the others are crashing when calling that function. 

Going deep in that function code I think it's crashing when it calls functions like ecs_query_iter/ecs_query_next on a query stored in a system. The tests where a query is manually built and those functions are used are working. 
Instead for the tests that use systems and call `ecs.progress`, which internally at some point uses  ecs_query_iter/ecs_query_next on a system query, they crash. Perhaps I messed up the way I represent the query inside a system?

Any help is welcome :) also I can't manage to setup a debugger with breakpoints so this investigation it's going super slow.

Closes https://github.com/zig-gamedev/zig-gamedev/issues/649


hazeycode:
cc @Srekel @btipling

Srekel:
If you are on windows I can help you set up VS code for debugging if you want. :)

Also it might be worth double checking our struct definitions so that they match with regards to defines such as FLECS_SANITIZE or whatever they are called.

giusdp:
@Srekel unfortunately I'm on mac! As soon as I can get back at it I'll double check the structs, and try again with the debugger.

libs/zflecs/src/zflecs.zig
```diff
@@ -2415,8 +2436,9 @@ pub fn SYSTEM(
...
     entity_desc.id = new_id(world);
     entity_desc.name = name;
-    entity_desc.add[0] = if (phase != 0) pair(EcsDependsOn, phase) else 0;
-    entity_desc.add[1] = phase;
+    const first = if (phase != 0) pair(EcsDependsOn, phase) else 0;
+    const second = phase;
+    entity_desc.add = &.{ first, second };
 
     system_desc.entity = entity_init(world, &entity_desc);
     _ = system_init(world, system_desc);
```
giusdp, line 2444: Looking at the execution with the debugger, it has a 0x0000000000000000 value for the `callback` field.
The thing is that the system_desc struct is correctly initialized and has a correct callback value, but the C code `system_init` receives an empty struct (with an empty callback). 

It might be an easy fix, I just don't understand why zig `system_desc_t` is ok but when it is passed to the C function it disappears.

Srekel: Just to confirm @giusdp is this something you get while running the tests? I'd like to help get 4.0 updated if I can :)

Srekel: Ok I've set up your fork and get the same crash. The test is identical to the official test as far as I can tell. I have to think there is something different with the system initialization compared in v4, which seems reasonable given that filters and such has changed a lot. 

I'm too tired to dig into it now but I can see two paths to find the answer: Either compare the C versio with the zig version and see whre they differentiate (it is quite different indeed) or diff flecs v3 and v4 to see how it changed. I'll see if I can dig into it more during the weekend.

giusdp: I didn't have much time to work on it unfortunately, @Srekel . I'm glad you already managed to get my fork working.

I tried to dig into it and also experimented with using `@cImport(@cInclude("flecs.h))` to see if the callback pointer was still disappearing, in that case it was not (although it was crashing later, but at least the pointer didn't turn into 0). 

I'm happy to pass you the torch and I hope you have immediately find the fix :)

candrewlee14: Did some debugging for you, looks like you've got some sizes wrong so the callback data wasn't showing up in the right place in memory.
`term_t`, for example, has converted some fields to use `i16` rather than `i32`.
Why is this not already cIncluding flecs.h to keep those in sync?

Srekel: Thanks, nicely spotted. Going through it to double check the rest. :)

Someone tell me if I'm wrong, but the reason for not using cInclude is that it allows for a more Zig-ish API.

Srekel: I've found a couple more issues. It gets past that test but now fails on the next instead. I'll continue going through it as I have time and energy. :) I'm not sure how to contribute best, @giusdp do you mind if I use my fork off of zig-gamedev and commit/PR from there, and mark you as a co-author?

giusdp: Sure no problem for me!

candrewlee14: > Thanks, nicely spotted. Going through it to double check the rest. :)
> 
> 
> 
> Someone tell me if I'm wrong, but the reason for not using cInclude is that it allows for a more Zig-ish API.

Do you have a specific example of this? It doesn't seem like the Zig API follows Zig's usual naming conventions

hazeycode: I thin kit's worth mentioning here that for other libs where we have headers to hand we can assert ABI compatibility. For example: https://github.com/zig-gamedev/zig-gamedev/blob/97586a8f5d738564963d3fb127c00fa3cc35d191/libs/zgpu/src/wgpu.zig#L3 although this would be better as a compile-time assertion rather than done via tests.

Srekel: Oh that's a great idea, I'll see if I can do something similar for zflecs :)

PR #660, opened 2024-08-04
[zgui] Add Node Editor support

OndraVoves:
Add support for [ImGui-node-editor](https://github.com/thedmd/imgui-node-editor)


hazeycode:
Nice!

PR #663, opened 2024-08-05
zphysics: fixup definition of AABox / JPC_AABox

kcbanner:
JPH::AABox uses Vec3 which is defined as:

```
 union
 {
  Type     mValue;
  float     mF32[4];
 };
```


hazeycode:
Thanks

PR #664, opened 2024-08-06
[zgui] BREAKING: Upgrade ImGui + Test Engine to 1.91.0

OndraVoves:
- This PR is only upgrade == no new features.
- Small number of breaking changes (some renames and new arg to `DockSpaceOverViewport`) check `gui.zig` changes.

### DIFF
--- a/libs/zgui/src/gui.zig
+++ b/libs/zgui/src/gui.zig
@@ -585,7 +585,7 @@ pub const SliderFlags = packed struct(c_int) {
     logarithmic: bool = false,
     no_round_to_format: bool = false,
     no_input: bool = false,
-    wrap_around: bool  = false,
+    wrap_around: bool = false,
     _padding: u23 = 0,
 };
 //--------------------------------------------------------------------------------------------------
@@ -1119,7 +1119,7 @@ pub const StyleCol = enum(c_int) {
     table_border_light,
     table_row_bg,
     table_row_bg_alt,
-    text_link,  
+    text_link,
     text_selected_bg,
     drag_drop_target,
     nav_highlight,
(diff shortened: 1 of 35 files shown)


libs/zgui/src/gui.zig
```diff
@@ -563,14 +564,14 @@ pub const WindowFlags = packed struct(c_int) {
 
 pub const ChildFlags = packed struct(c_int) {
     border: bool = false,
-    no_move: bool = false,
```
a-day-old-bagel, line 566: I think this has broken some of the samples, and will need fixing.

I'm also not sure I understand what "replaced" the .no_move flag, if anything did replace it. .nav_flattened does not appear to do the same thing.

OndraVoves: Nice 👍 which samples you mean? because this is correct change => `No_move` does not exist in original imgui  [ImGuiChildFlags_](https://github.com/ocornut/imgui/blob/29fadad1939ab62adcc5cd9339db7a784de90120/imgui.h#L1104) and only exist in [ImGuiWindowFlags_](https://github.com/ocornut/imgui/blob/29fadad1939ab62adcc5cd9339db7a784de90120/imgui.h#L1055). The problem is probably somewhere else

a-day-old-bagel: Ah ok I think you're right - why was no_move in the child flags before? Was that a mistake before? Sorry I got confused. Thanks for the clarification.

OndraVoves: Yes 😄  mistake I made in this [MR](https://github.com/zig-gamedev/zig-gamedev/pull/537/files#diff-aa21114d8640759f2d4590e4acb73790ce6d2bd4179d121cc1e1d613e1b9e2c7R562)

PR #673, opened 2024-08-13
Update zflecs to 4.0.0

Srekel:
This is based on @giusdp's first PR ( https://github.com/zig-gamedev/zig-gamedev/pull/657 ), then with changes primarily by @candrewlee14, and some by me. The tests now run fully. I have not yet tested this in for example Tides because I have a whole util layer that will need to be massaged in order to be compatible too. 

Relevant doc: https://www.flecs.dev/flecs/md_docs_2MigrationGuide.html


Srekel:
Btw for anyone wandering here now or in the future, here is my VS Code launch setup to debug the tests from Code.
```
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "(Windows) Launch",
            "type": "cppvsdbg",
            "request": "launch",
            "program": "${workspaceFolder}/zig-out/bin/zflecs-tests.exe",
            "console": "integratedTerminal",
            "cwd": "${workspaceFolder}",
            // "stopAtEntry": false,
            "args": [],
            "preLaunchTask": "build"
        }
    ]
}

```
And my tasks.json
```
{
    "version": "2.0.0",
    "tasks": [
        {
            "label": "build",
            "type": "shell",
            "command": "zig build",
            "group": "build",
            "problemMatcher": [
                "$gcc"
            ],
            "presentation": {
                "clear": true,
            }
        },
    ]
}
```

libs/zflecs/src/zflecs.zig
```diff
@@ -2534,6 +2565,8 @@ pub fn ADD_SYSTEM_WITH_FILTERS(
     filters: []const term_t,
 ) void {
     var desc = SYSTEM_DESC_WITH_FILTERS(fn_system, filters);
+    // std.json.stringify(desc, .{}, std.io.getStdOut().writer()) catch unreachable;
+    std.debug.print("lol234 {any}", .{desc});
```
candrewlee14, line 2569: Can we remove these lines?

Srekel: Oops!

Srekel: Fixed! Thanks for noticing! @hazeycode can we do a squash and erase my failboating? :)

libs/zflecs/src/tests.zig
```diff
@@ -16,6 +17,63 @@ test {
...
+        }
+        const ZigType = @field(@This(), decl.name);
+        if (@TypeOf(ZigType) != type) {
+            continue;
+        }
+        if (comptime std.meta.activeTag(@typeInfo(ZigType)) == .Struct and
+            @typeInfo(ZigType).Struct.layout == .@"extern")
+        {
+            const flecs_name = if (comptime std.mem.startsWith(u8, decl.name, "Ecs")) decl.name else "ecs_" ++ decl.name;
+            // const flecs_name = "ecs_" ++ decl.name;
```
hazeycode, line 52: Minor nit: delete old commented out code?

libs/zflecs/src/zflecs.zig
```diff
@@ -2860,7 +2891,6 @@ test {
 comptime {
     _ = @import("tests.zig");
```
hazeycode, line 2892: What's this doing here? Should we delete it considering that evaluating the definitions in tests.zig is wasted effort. Also, is this causing circular references? And can we drop the branch eval quota in the abi compatibility test if this is removed?

libs/zflecs/src/zflecs.zig
hazeycode: Of course ;)

libs/zflecs/src/zflecs.zig
Srekel: This was added by @michal-z 1½ years ago, I guess when initially working on the bindings. I tried removing it and then switching root source file for the test to:
`        .root_source_file = b.path("src/tests.zig"), `

And it seems to work.

And yes, I could now remove the 12325 workaround (I assume that's what you meant) and it still compiles and runs all tests successfully.

@michal-z remember why it was done like this? Any objection to me removing it? :)

michal-z: I don't remember, should be safe to remove.

Srekel: Fixed!

Srekel:
I noted that 4.0.1 just dropped, it seems to have a bunch of good fixes and indeed too many to just chuck on to this PR, so I will look into updating to that this weekend. 

https://github.com/SanderMertens/flecs/releases/tag/v4.0.1

PR #675, opened 2024-08-17
zphysics: Exposed Shape.GetLocalBounds and Shape.CastRay

a-day-old-bagel:
Also moved DebugRenderer.AABox to simply AABox, since it should have been independent anyway to reflect Jolt's architecture, and needs to be used for GetLocalBounds.

AABox also should have been using Reals instead of floats, so that is also fixed.


a-day-old-bagel:
hm, I'm not sure why the build is failing on here. It looks like all the toJpc and toJph macros are failing to build in this environment? That's odd. Maybe the error messages are just a side effect of what's actually wrong?

a-day-old-bagel:
I'm going to close this until I figure it out. Give me a bit.

PR #676, opened 2024-08-17
zflecs: Update to flecs 4.0.1

Srekel:
All 11 tests run successfully, haven't done any additional testing.

### DIFF
--- a/libs/zflecs/src/zflecs.zig
+++ b/libs/zflecs/src/zflecs.zig
@@ -2,7 +2,10 @@ const std = @import("std");
 const assert = std.debug.assert;
 const builtin = @import("builtin");
 
-pub const flecs_version = "4.0.0";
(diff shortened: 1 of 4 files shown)


libs/zflecs/src/zflecs.zig
```diff
@@ -2,7 +2,10 @@ const std = @import("std");
 const assert = std.debug.assert;
 const builtin = @import("builtin");
 
-pub const flecs_version = "4.0.0";
+pub const flecs_version = "4.0.1";
```
OndraVoves, line 5: Maybe a better use  [std.SemanticVersion](https://ziglang.org/documentation/0.13.0/std/#std.SemanticVersion) ;)

Srekel: Fixed :)

PR #677, opened 2024-08-17
zphysics: Expose Shape.GetLocalBounds and Shape.CastRay, move AABox to comply with Jolt structure.

a-day-old-bagel:
DebugRenderer.AABox should have simply been AABox, to reflect Jolt's architecture, and needs to be used for GetLocalBounds.

AABox also should have been using Reals instead of floats, so that is also fixed.

Hopefully this time the ci will like it.


hazeycode:
Thanks

PR #678, opened 2024-08-21
zglfw: Add glfw.setErrorCallback

### LINKED ISSUE #621: [zglfw] Add bindings for `glfwSetErrorCallback()`
I am switching from mach_glfw to zglfw because of Mach's zig version restrictions which stop ZLS from working. Anyway, I've noticed that apparently, there is no equivalent to `glfwSetErrorCallback`. Am I wrong? If not, I'd like to have such a binding.

Thanks in advance!

JamesHoltom:
Adds a binding for the `glfwSetErrorCallback` function.

Closes #621.


hazeycode:
Thanks!

PR #679, opened 2024-08-22
add zphysics instanced cubes d3d12 sample

Pyrolistical:
See https://github.com/zig-gamedev/zig-gamedev/blob/2fab51766f6525a44f84b679c69bcf0fbafc7af3/samples/zphysics_instanced_cubes_d3d12/README.md

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -121,6 +121,7 @@ const samples_windows_linux = struct {
     //pub const simple_raytracer = @import("samples/simple_raytracer/build.zig");
     pub const textured_quad = @import("samples/textured_quad/build.zig");
     pub const triangle = @import("samples/triangle/build.zig");
+    pub const zphysics_instanced_cubes_d3d12 = @import("samples/zphysics_instanced_cubes_d3d12/build.zig");
 };
 
 const samples_cross_platform = struct {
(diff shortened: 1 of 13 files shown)


hazeycode:
Nice! This works great on native Windows. However, the build is failing for some reason when cross-compiling from Linux. I did notice also that although the window can be resized, the frame buffer size is unchanged.

Pyrolistical:
I'll fix the framebuffer failing to resize

Pyrolistical:
> However, the build is failing when cross-compiling from Linux

I'll get Ubuntu installed and repo

hazeycode:
> > However, the build is failing when cross-compiling from Linux
> 
> If we care about this, should this not be an github action?

Yes, good point. I will update the CI script.

Pyrolistical:
Thanks for adding the CI, but what the heck is `zwin320`

hazeycode:
I think this is to do with target mismatch. In triangle/build.zig, the zpix target option is specified but it was using the default here.

It turns out that zpix does not even need to depend on zwin32 at all. I have removed the dependency. Can you try reintroducing zpix to your sample app and rebasing on main branch?

hazeycode:
Merged this here in https://github.com/zig-gamedev/zig-gamedev/pull/684 (GitHub wouldn't let me edit this PR)

Thanks!

PR #682, opened 2024-08-23
[zgui] Add isRectVisible

OndraVoves:


hazeycode:
Thanks.

Btw, incrementing the minor version is only required for breaking changes. For additions and other non-breaking changes, incrementing the patch version is sufficient.

(I've been sloppy with version numbers, often not bumping them, this will change when libs are in their own repo.)

PR #685, opened 2024-08-26
[Breaking] Fold zwin32, zxaudio2 and zd3d12 into zwindows

hazeycode:
These libs are almost always used together. Why not just put them together into an uber windows sdk Zig package?

Checklist
- [x] zwin32 => zwindows
- [x] zxaudio2 => zwindows/zxaudio2
- [x] zd3d12 => zwindows/zd3d12
- [x] Update README

Changes to build.zig usage:
```zig
    const zwindows_dependency = b.dependency("zwindows", .{
        // If using the optional helper libs, here are some good settings to start with:
        // (otherwise all default to `false`)
        .zxaudio2_debug_layers = (builtin.mode == .Debug),
        .zd3d12_debug_layers = (builtin.mode == .Debug),
        .zd3d12_gbv = b.option(bool, "zd3d12_gbv", "Enable GPU-Based Validation") orelse false,
    });
    
    // Import the Windows API bindings
    exe.root_module.addImport("zwindows", zwindows_dependency.module("zwindows"));

    // Import the optional zxaudio2 helper library
    exe.root_module.addImport("zxaudio2", zwindows_dependency.module("zxaudio2"));
    
    // Import the optional zd3d12 helper library
    exe.root_module.addImport("zd3d12", zwindows_dependency.module("zd3d12"));
```

Other changes:
- `w32.zig` renamed to `bindings/windows.zig` as this is now serving as an extension of `std.os.windows` (parts of which will be upstreamed or deprecated).
```zig
const zwindows = @import("zwindows");
const HWND = zwindows.windows.HWND;
const hrPanic = zwindows.windows.hrPanic;
const hrPanicOnFail = zwindows.hrPanicOnFail;
const d3d12 = zwindows.d3d12;
// etc
```
- `zd3d12_enable_d2d` build option deleted. The user can just not call the d2d fns if they don't want it.
- `zd3d12_upload_heap_capacity` build option deleted. Now capacity can be set at runtime when calling `zd3d12.GraphicsContext.init` who's function signature has been changed to take an args struct.
- Renamed top-level build options for consistency.


Srekel:
Will zxaudio2 and z3d12 be optional submodules to zwindows, or how will it work?

hazeycode:
> Will zxaudio2 and z3d12 be optional submodules to zwindows, or how will it work? 

Yes they could be zig modules provided by zwindows dependency. I will update the PR description later to explain better.

Srekel:
Just in case you're not aware, there's a proposal to remove usingnamespace, so it *may* be worth avoiding. Seems like it could go either way at the moment.

https://github.com/ziglang/zig/issues/20663

hazeycode:
I've removed the `usingnamespace`s except for the one that pulls in `std.os.windows`. This can be removed later without affecting the public API.

PR description has been updated to reflect the changes.

hazeycode:
Any objections to me merging this as is?

Pyrolistical (approved):

PR #691, opened 2024-09-02
zgui doesn't cover imgui API, missing drag functionality

SorenSaket:
I was porting another application using Imgui and realized IsMouseDragging() was missing. 

Since the bindings are hand written I'm wondering if there's other discrepancies. I fixed some others I need for now. Also there's overloads which aren't added.

We should consider non hand written in the future for robustness.

I don't know is it's okay to do this, but I changed the return type from MouseClickedCount to u32 since the value doesn't go below 0.

### DIFF
--- a/libs/zgui/src/gui.zig
+++ b/libs/zgui/src/gui.zig
@@ -3266,8 +3266,12 @@ pub const MouseButton = enum(u32) {
 pub const isMouseDown = zguiIsMouseDown;
 /// `pub fn isMouseClicked(mouse_button: MouseButton) bool`
 pub const isMouseClicked = zguiIsMouseClicked;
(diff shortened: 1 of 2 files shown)


hazeycode:
I agree, because Dear Imgui is ever-evolving, it would make sense to generate the bindings instead of hand-wrangling them.

Srekel:
In theory I suppose we could use https://github.com/lassade/c2z (I've used it for other libraries) but there's definitely a quality difference in the bindings (that could definitely be improved by improving c2z). One could write something custom but that's a big time/tech investment. Or we could use for example c2z to generate basic bindings and then have a nicer API on top. 

The current bindings, while incomplete, are pretty darn nice!

PR #702, opened 2024-09-09
zphysics: Fixed types for double precision support

a-day-old-bagel:
All tests should now pass with both single and double-precision floats set in Jolt.


a-day-old-bagel:
Hold up, I should add one more fix to this.

a-day-old-bagel:
Ok, now it should be good to go, I believe.

Most of the changes were to fix structs that had used Reals when Jolt was actually just using floats in that spot.

A separate struct for single-precision ray casts was added, and a couple small associated tests added into the basic body test.

PR #703, opened 2024-09-10
zgui: add support for FreeType font rendering

okvik:
This adds simple build support to have Dear ImGui use the system's FreeType library for nicer-looking fonts.

Bundled FreeType should not be too hard to provide either, but I'd like to see first with the maintainers whether this is something that you'd be interested to include in the project.

### DIFF
--- a/libs/zgui/build.zig
+++ b/libs/zgui/build.zig
@@ -118,13 +118,14 @@ pub fn build(b: *std.Build) void {
     });
 
     if (options.with_freetype) {
-        imgui.addCSourceFile(.{
-            .file = b.path("libs/imgui/misc/freetype/imgui_freetype.cpp"),
-            .flags = cflags,
-        });
-        imgui.defineCMacro("IMGUI_ENABLE_FREETYPE", "1");
-        const freetype = b.dependency("freetype", .{});
-        imgui.linkLibrary(freetype.artifact("freetype"));
+        if (b.lazyDependency("freetype", .{})) |freetype| {
+            imgui.addCSourceFile(.{
+                .file = b.path("libs/imgui/misc/freetype/imgui_freetype.cpp"),
+                .flags = cflags,
+            });
+            imgui.defineCMacro("IMGUI_ENABLE_FREETYPE", "1");
+            imgui.linkLibrary(freetype.artifact("freetype"));
+        }
     }
 
     if (options.use_wchar32) {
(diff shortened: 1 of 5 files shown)


okvik:
Turns out someone has already ziggified FreeType: https://github.com/hexops/freetype, so it is quite trivial to have it built from source instead.

okvik:
The rest of the functionality provided by ImGui FreeType integration -- support for grabbing the builder interface and setting a custom allocator on it -- mostly depends on zgui exposing the ImFontAtlas functionality. However, that's something that warrants a separate PR. I'll work on that next if / when the FreeType stuff gets merged.

libs/zgui/build.zig
```diff
@@ -112,6 +117,16 @@ pub fn build(b: *std.Build) void {
         .flags = cflags,
     });
 
+    if (options.with_freetype) {
+        imgui.addCSourceFile(.{
+            .file = b.path("libs/imgui/misc/freetype/imgui_freetype.cpp"),
+            .flags = cflags,
+        });
+        imgui.defineCMacro("IMGUI_ENABLE_FREETYPE", "1");
+        const freetype = b.dependency("freetype", .{});
```
hazeycode, line 126: `lazyDependency` would be better here so that freetype is only fetched if the option is enabled.
This reminded me to make this commit: https://github.com/zig-gamedev/zig-gamedev/commit/cf15279a372dcb8d0482bcb6b89851b562c122ed and start making use of lazy dependencies.

hazeycode (changes requested):

hazeycode:
Thanks for this. I have left a couple of small feedback requests.

libs/zgui/src/gui.zig
```diff
@@ -154,6 +154,23 @@ pub const ConfigFlags = packed struct(c_int) {
     _padding: u10 = 0,
 };
 
+pub const FontBuilderFlags = if (@import("zgui_options").with_freetype == false)
+    c_uint
+else
+    packed struct(c_uint) {
```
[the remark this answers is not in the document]
okvik: Agreed. It's not necessary.

libs/zgui/build.zig
okvik: Yea. I made a local commit that uses lazy dependencies for exactly the reason you mention. But since all the other dependencies were still eager I decided against being the first to break the ice :P

okvik:
> Thanks for this. I have left a couple of small feedback requests.

Thanks for the review. I'll make the requested changes soon.

okvik:
Done. Would you like me to squash the commits before merging?

hazeycode (approved):

PR #704, opened 2024-09-11
zgui: add ImGui default colors style API

okvik:
This adds the `ImGui::StyleColors{Dark,Light,Classic}` API to the `zgui.Style` struct.

Compressing these functions to a single function with a style variant argument and naming it `Style.setDefaultColors` is a choice that's up for debate. It seems kinda nice to me, but I'm open to changing it to whatever if people disagree.

### DIFF
--- a/libs/zgui/src/gui.zig
+++ b/libs/zgui/src/gui.zig
@@ -1062,6 +1062,22 @@ pub const Style = extern struct {
     pub const scaleAllSizes = zguiStyle_ScaleAllSizes;
     extern fn zguiStyle_ScaleAllSizes(style: *Style, scale_factor: f32) void;
 
+    pub const DefaultColors = enum {
+        dark,
+        light,
+        classic,
+    };
+    pub fn setDefaultColors(style: *Style, variant: DefaultColors) void {
+        switch (variant) {
+            .dark => zguiStyleColorsDark(style),
+            .light => zguiStyleColorsLight(style),
+            .classic => zguiStyleColorsClassic(style),
+        }
+    }
+    extern fn zguiStyleColorsDark(style: *Style) void;
+    extern fn zguiStyleColorsLight(style: *Style) void;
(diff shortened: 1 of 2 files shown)


hazeycode (changes requested):
This is nitpicky but I think we should **not** call these "default colours" and rather name `DefaultColours` as `StyleColoursBuiltin` and instead of `setDefaultColors` we should have `styleColorsDark`, `styleColorsLight` and `styleColorsClassic` as well as `styleColorsBuiltin`. This would be closer to the underlying API and offer the least surprises for users.

okvik:
Alright, will do. Though I don't think I understand from what you wrote what exactly you'd like the API to be.

Keeping up with the upstream API there's two options:

1. Style.styleColorsDark() or Light() or Classic()
2. Style.styleColors(.dark) or .light or .classic

hazeycode:
I had misread things and my previous message was confused and confusing. Sorry about that.

I think the core API here should be:
`styleColorsDark(&style)`, `styleColorsLight(&style)`, `styleColoursClassic(&style)` defined in the top level namespace.

In addition, as "member functions" of `Style` we can have:
`style.setColorsDark()`, `style.setColorsLight()`, `style.setColoursClassic()`
which can be defined as equal to the top-level definitions.

We do [something similar in zsdl.](https://github.com/zig-gamedev/zig-gamedev/blob/fbf96714ab6876646a29824a3b4ba79e34906a6d/libs/zsdl/src/sdl2.zig#L200). And I think this a nice way to offer both plain C-style bindings and an embellished "zigified" API that the user can choose at their discretion.

And I think there is value in embellishing the API further with an enum variant as you have done in your original changeset. `style.setColorsBuiltin(.light)`.

okvik:
Done. Would you like me to squash the changes or leave the history as-is?

hazeycode:
Thank you ❤️
And that's ok, I can squash on merge.

hazeycode (approved):

PR #706, opened 2024-09-16
zphysics: Fixed DebugRenderer for double precision builds, created explicit RMatrix class

a-day-old-bagel:
Also updated the monolith demo to reflect the changes.

Before, while fixing the double precision builds, I missed something in DebugRenderer that wasn't correct. I'm basically just adding this fix.

DebugRenderer should now work in both precision modes, instead of just in single precision mode.


hazeycode:
Thanks for following up ❤️

PR #708, opened 2024-09-21
[breaking] refactor zsdl prebuilt libs

hazeycode:
Split `sdl2-prebuilt` into `sdl2-prebuilt-macos`, `sdl2-prebuilt-x86_64-linux-gnu` and `sdl2-prebuilt-x86_64-windows-gnu` and consume them lazily.


hazeycode:
FYI @gegogi

PR #715, opened 2024-10-05
Some fix for 0.14.0-dev.1710+8ee52f99c

OndraVoves:


libs/zflecs/src/zflecs.zig
```diff
@@ -841,7 +841,7 @@ pub const entity_desc_t = extern struct {
     symbol: ?[*:0]const u8 = null,
     use_low_id: bool = false,
     add: ?[*:0]const id_t = null,
-    set: ?[*:.{}]const value_t = null,
+    set: ?[*]const value_t = null, // TODO: non-scalar terminator are deprecated in new zig
```
OndraVoves, line 844: Breaking change => user must set terminator by self

PR #718, opened 2024-10-11
Upgrade Zig to new Mach nominated ver 0.14.0-dev.1911+3bf89f55c

hazeycode:


OndraVoves:
btw, nominations are still in progress. There was a bug in ZIG, it has been fixed, but it will probably be a different version than 0.14.0-dev.1710+8ee52f99c.

OndraVoves:
@hazeycode  I try compile this branch on my Mac and its compile without any problem. In my project I use **zglfw**, **zpool**, **zjobs**, **ztracy**, **zgui**, **zflecs** and everything works fine with new nominated zig.

hazeycode:
I can confirm this branch is building locally for me on an Intel Mac running macOS 13.5 and on an Apple silicon Mac running macOS 14.5. I wonder what's different with the CI runner environment.

hazeycode:
Other than macOS... Linux is working for me locally. For windows, I have pushed some more fixes, but `monolith` and `physics_test_wgpu` produce linker errors; note these are the only two samples that use both Dawn (zgpu) and Jolt (zphysics) together.

a-day-old-bagel:
I notice that zgpu and zphysics seem to be the only two libraries that are building with the `-std=c++17` argument.  And the linker errors are complaining about duplicate definitions for `std::__1::bad_function_call` and stuff.

I wonder if those two facts are related.

a-day-old-bagel:
I've been trying to fix this, but it's proving quite difficult.

I *believe* that the issue is that dawn.lib, which is the precompiled lib we are just pulling from github as described in build.zig.zon, is exporting a bunch of these C++ function definitions that are conflicting with the definitions given elsewhere. I've tried all sorts of things to try to get zig to treat those definitions as "weakly linked" so that other definitions can override them without producing an error. So far nothing has helped.

I don't know what changed in zig to produce this behavior, but could it also be that the dawn.lib that was compiled for Windows could be recompiled in a way that doesn't export some of those symbols?

PR #719, opened 2024-10-11
update ztracy readme instruction

JustinBraben:
Updating ztracy readme. Hopefully makes it easier to get other projects up and running with ztracy!

If the `const options` is too much I can rework it.


hazeycode (approved):
Thanks!

Issue #714, opened 2024-10-04
ztracy Always Enabled


jcalabro:
Hey there! Even when [enable_ztracy](https://github.com/zig-gamedev/zig-gamedev/blob/d96ecc993bcfc4461f44d9432589dc9273952a78/libs/ztracy/build.zig#L8C10-L12) is false, tracy is enabled and the application is listening on port 8080.

Take the procedural mesh sample application for instance:

![image](https://github.com/user-attachments/assets/90a2433a-2aa0-4cf7-b743-66fea75fec36)

You can see that the application is listening on 8086 and tracy can find it without issue even though `enable_ztracy` is false.

I think we want to change [this line](https://github.com/zig-gamedev/zig-gamedev/blob/d96ecc993bcfc4461f44d9432589dc9273952a78/libs/ztracy/build.zig#L50) to:

```zig
if (options.enable_ztracy) "-DTRACY_ENABLE" else "",
```

Which appears to solve the issue locally. Let me know what you think!
\- Jim

hazeycode:
Fixed by https://github.com/zig-gamedev/zig-gamedev/pull/722

jcalabro:
Thanks all! :heart:

PR #722, opened 2024-10-21
Add option for shared library. Update -DTRACY_ENABLE option flag

### LINKED ISSUE #693: [ztracy] Add support to build as shared libray
Currently ztracy can only be built as a static library, but tracy already supports being built as a shared library, it would be a nice enhancement to add an option to pick library type.

https://github.com/ziglang/zig/issues/19341 might impact the implementation of this.

JustinBraben:
Closes #693 and #714 

Added option for shared library. Pass `-Dshared=true` to build ztracy as a shared library.
Now passing `-DTRACY_ENABLE` only if the option was specified.


hazeycode:
Thanks!

Issue #622, opened 2024-06-12
Update to latest jolt physics


Pyrolistical:
Jolt physics was last updated at version 3.0.1. Latest release is 5.0.0

[Release notes](https://jrouwe.github.io/JoltPhysics/md__docs_2_release_notes.html)

Breaking changes:
[Changes between v3.0.1 and v4.0.0](https://jrouwe.github.io/JoltPhysics/md__docs_2_a_p_i_changes.html#autotoc_md31)
[Changes between v4.0.0 and v5.0.0](https://jrouwe.github.io/JoltPhysics/md__docs_2_a_p_i_changes.html#autotoc_md25)

Pyrolistical:
I tried to update to 5.0.0 and it fails to compile since JoltC needs to be resynced.

### Sample of JoltC issues
1. [`BodyManager.mNumActiveBodies` and `BodyManager.mActiveBodies` becames an array](https://github.com/jrouwe/JoltPhysics/pull/644/files#diff-9934629525caac92050493731fe814b7a7433dfaee6dc82f3676314b4b90d7cdL304-L307)
2. `JPC_BodyCreationSettings` out of sync
3. `JPC_ContactSettings` out of sync
4. `JPC_SHAPE_TYPE_USER[1-4]` out of sync
5. `JPC_SHAPE_SUB_TYPE_USER[1-8]` out of sync
6. `JPC_SHAPE_SUB_TYPE_USER_CONVEX1[1-2]` out of sync

pfeodrippe:
@Pyrolistical @hazeycode, I am working on a Clojure game framework and I’m using zphysics, it’s pretty awesome, thanks for this work!!

If you want some reference, I have been able to use the latest version of Jolt (as of today), see the changes at https://github.com/zig-gamedev/zig-gamedev/commit/9ab61bd6f54e1967f140205df5cfe27d4ca02aee (I have deleted the src zig file as there were expected size differences for some structs and I am not using it).

Pyrolistical:
taking a look at this now @pfeodrippe

a-day-old-bagel:
I can give this a shot, if nobody else is actively working on it.

Latest release is now 5.1.0, it appears.

a-day-old-bagel:
Ok, I got the new version compiling, but I'm still tracking down a few memory-related crashes and test fails, probably due to some structure I missed and still need to update to reflect Jolt's changes.

But basically, this looks quite doable and near completion.

[...]

PR #723, opened 2024-10-21
Jolt Physics updated to 5.1.0 release

a-day-old-bagel:
In response to #622, as well as my own desire to use newer releases of Jolt.

This is an assorted bag of changes. I've only had access to a Windows machine over the past few days, so we may want to get some of you to also test this on your systems before considering it a mature PR, if we're not confident in the results of the CI alone.

I've probably missed something somewhere, but after testing all the things I can think of, I haven't found any more changes that need to take place on my own.

Part of the upgrade involved implementing a realloc function for Jolt to call, and I've just put in a naive approach that might need a second pair of eyes. I never used to use realloc much in C, so I just did it how I thought most logical, and so far all the tests run, and my application also runs, and I can confirm Jolt 5.1.0 is calling realloc quite a bit. And sometimes, it does successfully resize an allocation, so whatever benefit Jolt's author thought might be available via resizing allocations should be possible through our zphysics allocator implementation.

All I have done is maintain our currently exposed API feature set, and gotten it to run on top of 5.1.0, to be clear. Jolt has added a bunch of soft body stuff, but I haven't done any additional work to actually expose any of that. Only to bring our existing exposed API bits into sync.

Please let me know if there's anything here we'd prefer to do differently.


michal-z:
For `realloc()` you could use `Allocator.realloc()` which would simplify your implementation. `zmesh` does similar thing here:
https://github.com/zig-gamedev/zig-gamedev/blob/main/libs/zmesh/src/memory.zig#L71

a-day-old-bagel:
Ok, I switched to using the approach from zmesh, and it did seem to work just the same, so I must have just bumbled around in the wrong direction before when trying those methods.

hazeycode:
Tested `physics_test_wgpu` and `monolith` sample apps with these changes on an Intel Mac running macOS 13.5. Everything appears to be working as before. Changes look good too. Along with passing tests, that's enough for me to happily merge this. Missing APIs can be added later. If there are any unearthed errors we should expand the coverage of the tests and/or the sample app repertoire.  

Thank you for this. It looks like it took quite a lot of effort.

hazeycode (approved):
Tested `physics_test_wgpu` and `monolith` sample apps with these changes on an Intel Mac running macOS 13.5. Everything appears to be working as before. Changes look good too. Along with passing tests, that's enough for me to happily merge this. Missing APIs can be added later. If there are any unearthed errors we should expand the coverage of the tests and/or the sample app repertoire.  

Thank you for this. It looks like it took quite a lot of effort.

PR #668, opened 2024-08-09
Migrate all the libs

hazeycode:
All the libs to each their own git repository...

- [x] [system_sdk](https://github.com/zig-gamedev/system_sdk)
- [x] [zaudio](https://github.com/zig-gamedev/zaudio)
- [x] [zbullet](https://github.com/zig-gamedev/zbullet)
- [x] [zemscripten](https://github.com/zig-gamedev/zemscripten)
- [x] [zflecs](https://github.com/zig-gamedev/zflecs)
- [x] [zglfw](https://github.com/zig-gamedev/zglfw)
- [x] [zgpu](https://github.com/zig-gamedev/zgpu)
- [x] [zgui](https://github.com/zig-gamedev/zgui)
- [x] [zjobs](https://github.com/zig-gamedev/zjobs)
- [x] [zmath](https://github.com/zig-gamedev/zmath)
- [x] [zmesh](https://github.com/zig-gamedev/zmesh)
- [x] [znoise](https://github.com/zig-gamedev/znoise)
- [x] [zopengl](https://github.com/zig-gamedev/zopengl)
- [x] [zopenvr](https://github.com/zig-gamedev/zopenvr)
- [x] [zphysics](https://github.com/zig-gamedev/zphysics)
- [x] [zpix](https://github.com/zig-gamedev/zpix)
- [x] [zpool](https://github.com/zig-gamedev/zpool)
- [x] [zsdl](https://github.com/zig-gamedev/zsdl)
- [x] [zstbi](https://github.com/zig-gamedev/zstbi)
- [x] [ztracy](https://github.com/zig-gamedev/ztracy)
- [x] [zwindows](https://github.com/zig-gamedev/zwin32) (https://github.com/zig-gamedev/zig-gamedev/pull/685)

PR #731, opened 2024-11-04
zphysics: support for shape serialization

kcbanner:
Here are some proposed changes to add support for binary serialization of shapes.

- Add `StreamOut` and `StreamIn` interfaces
- Add implementations of `StreamOut` / `StreamIn` that use `AnyWriter` / `AnyReader`
- Add `Shape::saveBinaryState` and `Shape::RestoreFromBinaryState`

Opening this as a draft as I may be adding support for `SaveMaterialState` and `SaveSubShapeState` depending on if my use case requires it or not. I'm avoiding it for now as the map data structures may be tricky to easily map between C/C++.


hazeycode:
Looks good to me.
Note I am about to move zphysics to its own lib and complete the [lib migration](https://github.com/zig-gamedev/zig-gamedev/pull/668).

kcbanner:
Cool, I'll hold off until the migration and PR there, as I have some more changes for this patch - it turns out I do need to use SaveWithChildren / RestoreWithChildren in my use case so I need to add those as well.

kcbanner:
Moved here https://github.com/zig-gamedev/zphysics/pull/1

PR #748, opened 2025-03-27
Fix build.zig.zon and dev.build.zig.zon to be compatible with Zig 0.14.0

steepcloud:
- Updated `.name=.zig_gamedev` for Zig 0.14.0 compatibility

- The current setup works with 0.14.0-dev.2577 but not with the released 0.14.0


hazeycode:
See https://github.com/zig-gamedev/zig-gamedev/pull/750

